PR #28, opened 2021-09-05
fix mssql connection provider

Idokah:
Add .then() to pool.connect() while connecting to mssql because the function connect() throwing if the connection failed.
Change the cleanup function when connection is failed because it's throwing when we try to close failed pool in mssql.

### DIFF
--- a/packages/external-db-mssql/lib/connection_provider.js
+++ b/packages/external-db-mssql/lib/connection_provider.js
@@ -22,6 +22,14 @@ const extraOptions = () => {
     }
 }
 
+const notConnectedPool = (pool, err) => {
+    return {
+        ...pool,
+        query: async () => { throw err },
+        request: async () => { throw err },
+        connect: async () => { return await pool.connect() }
+    }
+}
 
 const init = async (cfg, _poolOptions) => {
     const config = {
@@ -39,8 +47,14 @@ const init = async (cfg, _poolOptions) => {
     const poolOptions = _poolOptions || {}
 
     const _pool = new ConnectionPool(Object.assign({}, config, extraOptions(), poolOptions))
-
-    const {pool, cleanup} = await _pool.connect().then((res) => {return {pool: res, cleanup: async () => await pool.close() }}).catch((e) => { return {pool:_pool, cleanup: () => { }} })
+    const { pool, cleanup } = await _pool.connect().then((res) => {
+        return { pool: res, cleanup: async () => await pool.close() }
+    }).catch((e) => {
+        return {
+            pool: notConnectedPool(_pool, e),
+            cleanup: () => { }
+        }
+    })
     const databaseOperations = new DatabaseOperations(pool)
 
     const filterParser = new FilterParser()
--- a/packages/external-db-mssql/lib/connection_provider.js
+++ b/packages/external-db-mssql/lib/connection_provider.js
@@ -1,6 +1,6 @@
 const { ConnectionPool } = require('mssql')
 const SchemaProvider = require('./mssql_schema_provider')
-const DataProvider  = require('./mssql_data_provider')
+const DataProvider = require('./mssql_data_provider')
 const FilterParser = require('./sql_filter_transformer')
 const DatabaseOperations = require('./mssql_operations')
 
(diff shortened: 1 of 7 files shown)


packages/external-db-mssql/lib/connection_provider.js
```diff
@@ -39,15 +39,15 @@ const init = async (cfg, _poolOptions) => {
     const poolOptions = _poolOptions || {}
 
     const _pool = new ConnectionPool(Object.assign({}, config, extraOptions(), poolOptions))
-    const pool = await _pool.connect()
 
+    const {pool, cleanup} = await _pool.connect().then((res) => {return {pool: res, cleanup: async () => await pool.close() }}).catch((e) => { return {pool:_pool, cleanup: () => { }} })
```
noam-almog, line 43: 1. line cannot be that long, please add some new lines next time.
2. i don't understand what happens in the case it's not connecting
3. why do we return cleanup and never use it later on ?

packages/external-db-mssql/lib/mssql_operations.js
```diff
@@ -9,16 +8,13 @@ class DatabaseOperations {
...
-                case 'ELOGIN':
-                    return { valid: false, error: new DbConnectionError(`Access to database denied - probably wrong credentials, sql message: ${err.message}`) }
-                case 'ESOCKET':
-                    return { valid: false, error: new DbConnectionError(`Access to database denied - host is unavailable, sql message: ${err.message}`) }
+            if (sql._connected) {
+                return await sql.query('SELECT 1')
+                    .then(() => { return { valid: true } })
+                    .catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e) } })
             }
+            return sql.connect().then(() => { return { valid: true } }).catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e) } })
```
noam-almog, line 16: why do we try to connect in this case ?

packages/external-db-mssql/lib/connection_provider.js
Idokah: 2. in the case it's not connecting, we get back the unconnected pool and empty cleanup, instead of throwing an error, later on we'll use the unconnected pool in the operations (try to connect again and then show the appropriate error), and when we try to send a request we'll get a "connection closed" error.
3. we use the cleanup - we returning it in the last line, I'm returning the empty cleanup because if we're trying to close the unconnected pool it's throwing an error.

noam-almog: operations cannot be a part of the connection flow, it just checks health of the pool

packages/external-db-mssql/lib/mssql_operations.js
Idokah: Similar to the query 'select 1' in other implementations, in mssql if we try to query a closed pool the only error we'll get is "connection is closed", and if we'll try to connect we'll get more informative error - like wrong credentials

Idokah: And in the first case (pool is connected), we're trying to query the connected pool and if it's worked, the pool is valid. else there's another error, although the pool is connected.

packages/external-db-mssql/lib/connection_provider.js
Idokah: After the connection is failed from any reason it's throwing an error, if I don't want the app will crash I need to return something from the conection_provider (connection, cleanup), so I thought it would make the most sense to return the unconnected pool, because when we're querying this pool is throwing appropriate error and not 'query is not a function' or something similar.

noam-almog: we can also create a synthetic pool that implements a query function (that will throw an error in case it's called)

Idokah: @noam-almog I created, tell me if that's what you meant

packages/external-db-mssql/lib/connection_provider.js
```diff
@@ -22,6 +22,14 @@ const extraOptions = () => {
     }
 }
 
+const notConnectedPool = (pool, err) => {
```
noam-almog, line 25: looks good, extract this to mssql_utils

packages/external-db-mssql/lib/connection_provider.js
noam-almog: looks good, fix my last warning and merge

packages/external-db-mssql/lib/connection_provider.js
Idokah: done

PR #29, opened 2021-09-09
fix aggregate api test - spanner

Idokah:
The aggregate api test failed because of the `givenCollection` function
`columns.map(async column => await axios.post('/schemas/column/add', {collectionName: name, column: column}, auth)) )` 

The function failed because cloud spanner can only execute one transaction at a time on a session, and we're trying to post column/add before the last transaction have done. 
so I've added retry to this post request, I would love to hear if you think there is a better solution or it's ok for now.

### DIFF
--- a/packages/velo-external-db/package.json
+++ b/packages/velo-external-db/package.json
@@ -17,6 +17,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "axios-retry": "^3.1.9",
     "compression": "^1.7.4",
     "ejs": "^3.1.6",
     "express": "^4.17.1",
@@ -39,7 +40,6 @@
     "jest-each": "^27.0.6",
     "jest-when": "^3.3.1",
     "test-commons": "^0.0.0",
-    "ts-jest": "^27.0.4",
-    "chance": "^1.1.7"
+    "ts-jest": "^27.0.4"
   }
 }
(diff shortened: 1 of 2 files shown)


noam-almog:
the test is failing because of this: 
```  ● Data API › Spanner › aggregate api without filter

    9 FAILED_PRECONDITION: Could not parse 8629 as a FLOAT64. Only the following string values are supported for FLOAT64: 'Infinity', 'NaN', and '-Infinity'. See https://cloud.google.com/spanner/docs/data-types#floating-point-type for more details

      at Object.callErrorFromStatus (../../node_modules/@grpc/grpc-js/src/call.ts:81:24)
      at Object.onReceiveStatus (../../node_modules/@grpc/grpc-js/src/client.ts:338:36)
      at ../../node_modules/@grpc/grpc-js/src/call-stream.ts:170:27
      at Object.onReceiveStatus (../../node_modules/grpc-gcp/src/index.ts:129:15)
      at InterceptingListenerImpl.onReceiveStatus (../../node_modules/@grpc/grpc-js/src/call-stream.ts:166:19)
      at Object.onReceiveStatus (../../node_modules/@grpc/grpc-js/src/client-interceptors.ts:426:34)
      at Object.onReceiveStatus (../../node_modules/@grpc/grpc-js/src/client-interceptors.ts:389:48)
      at ../../node_modules/@grpc/grpc-js/src/call-stream.ts:276:24```
fix it on another PR

PR #31, opened 2021-09-12
add temp fix for spanner float

Idokah:
This is a temp fix for the float problem with spanner.
We just make sure that we don't get a float when the column.subtype is 'decimal'.


Idokah:
If we want to really solve it, we need to insert the float number with Spanner.float(), after that the number becomes an object, and we need to parse it later.

PR #36, opened 2021-09-29
mongo operations, connection provider

Idokah:


packages/velo-external-db/test/storage/database_operation.spec.js
```diff
@@ -10,18 +10,30 @@ describe('Check Pool Connection', () => {
...
+        if (dbType === 'Mongo')
+            test('pool connection with wrong config will return object with valid equal to false and DbConnectionError.', async () => {
+                const dbOperation = (gen.randomObjectFromArray(env.driver.dbOperationWithMisconfiguredConfig))() 
+                // todo - move randomObjectFromArray to beforeEach if we keeping this and implement dbOperationWithMisconfiguredConfig for every driver.
+
+                const validateConnection = await dbOperation.validateConnection()
+
+                expect(validateConnection.valid).toBeFalsy()
+                expect(validateConnection.error).toBeInstanceOf(DbConnectionError)
+            })
```
Idokah, line 23: I thought to change the database operation tests to include only this test and the last one, with the valid pool (marked with //1, //2).
There are differences between the databases, for example in MongoDB the exception for wrong database/ /password/ username is the same ('Authentication failed'), and Firestore is very different and so on. 
so I thought that every driver will have an array with all the options to misconfigure the config, and then we'll choose one of them randomly (or all of them) and expect `DbConnectionError` instead of some kind of string.

PR #39, opened 2021-10-04
add columns to create method schemas/create

Idokah:


noam-almog:
i'm not sure why we need this ? we have add column api
plus, if we do decide to add this parameter, we need to add e2e tests for that.

PR #41, opened 2021-10-05
change mongo config reader to uri only

Idokah:


Idokah:
From what I have seen in all the places where there is a connection to Mongo it's with the URI, and also when you open the MongoDB the URI is produced (in atlas/azure at least)
I think it will be more convenient to ask only URI.

MXPOL:
Agree.
I have never seen a need to give both username & password to connect to MongoDB it's always a URI 🧐

PR #59, opened 2021-10-25
Airtable

Idokah:
app_data e2e pass (except aggregation),
app_schema e2e pass,
data_provider without aggregation and skip
schema_provider pass
db_operation with mock need to fix.

still need to refactor and go through it, so dont merge yet.


Idokah:
about aggregation- they didn't expose MIN, MAX, etc. so it's problematic to implement aggregation.
db_operation fixed.
about skip- for now i did temporary solution - didn't succeed to perform it through their client

PR #69, opened 2021-11-10
BigQuery support (WIP)

MXPOL:


noam-almog:
you should start working on the schema first, not sql filter or init, use the tests to implement each of the api's

PR #99, opened 2021-12-15
Api wrapper transform filter

Idokah:


noam-almog:
good pr, i'm missing some edge case tests like empty filter cases (code is there but no tests for it, fix it and then we can merge)

PR #111, opened 2021-12-22
Mysql with new filter, aggregation format

Idokah:
I've added some if conditions in `app.js` and in `data.js` to check whether the implementation support the new format or not. After we adjust all the implementations, we can remove those conditions.


packages/velo-external-db-core/lib/service/data.js
```diff
@@ -1,12 +1,17 @@
...
+    constructor(storage, schemaInformation, supportAdapterFormat) {
         this.storage = storage
         this.schemaInformation = schemaInformation
+        this.filterTransformer = supportAdapterFormat ? new FilterTransformer() : false
+        this.aggregationTransformer = supportAdapterFormat ? new AggregationTransformer(this.filterTransformer): false
     }
 
-    async find(collectionName, filter, sort, skip, limit) {
+    async find(collectionName, _filter, sort, skip, limit) {
+        const filter = this.filterTransformer ? this.filterTransformer.transform(_filter) : _filter
```
noam-almog, line 14: extract function

PR #128, opened 2021-12-27
supported schema operations

MXPOL:
This PR has a new function in schema_commons that returns the supported schema operations for each implementation,
I changed each schema provider to use that function.


MXPOL:
@noam-almog can you take a look at this PR? It looks like a lot of changes, but I did the same changes in each implementation. 
If it is ok, you can merge this PR

PR #129, opened 2021-12-27
validate query

Idokah:
I didn't add the validate function in data.js yet because I think that it will fail other tests sometimes, I'll change them first.

For now, the functions throws if the filter is filtering on not existed column, or filter with not allowed operator for the field data type.


Idokah:
I have not finished it yet, when I continue I need to create a new class that will both validate and transform (need to think what it will be called) and make sure it does not fail any data service or e2e tests due to the validation.

MXPOL:
Disabled some  `data.js` tests, it will be fixed in another PR that will refactor a little the test 
I want to dedicate a separate PR to this

PR #130, opened 2021-12-27
Test if operation supported

MXPOL:
This PR check if the implementation has the needed operation and if so it will run the tests and if not it will skip it


MXPOL:
Closing this PR, I will create new PR that uses new schema API call

PR #131, opened 2021-12-29
big query cleanup

noam-almog:


noam-almog:
@MXPOL please move schemaSupportedOperations to utils or constants on all modules after you merge

PR #139, opened 2022-01-05
fix for schema service tests

MXPOL:
The format of the column object have been changed, but the random column generator haven't, this pr fix it and generates the right column object

### DIFF
--- a/packages/test-commons/lib/gen.js
+++ b/packages/test-commons/lib/gen.js
@@ -57,9 +57,9 @@ const randomCollections = () => randomArrayOf( randomCollectionName )
 
 const randomFieldName = () => chance.word({ length: 5 })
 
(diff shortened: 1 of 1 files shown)


packages/test-commons/lib/gen.js
```diff
@@ -57,9 +57,9 @@ const randomCollections = () => randomArrayOf( randomCollectionName )
 
 const randomFieldName = () => chance.word({ length: 5 })
 
-const randomDbField = () => ( { name: chance.word(), type: chance.word(), subtype: chance.word(), isPrimary: chance.bool() } )
+const randomCollectionField = () => ({ field: chance.word(), type: randomWixDataType() })
```
noam-almog, line 60: why not just change name to field, why rename all of the methods ?

PR #142, opened 2022-01-09
Schema tests matchers

MXPOL:


packages/velo-external-db-core/lib/service/schema.spec.js
```diff
@@ -4,6 +4,10 @@ const driver = require('../../test/drivers/schema_provider_test_support')
 const schema = require('../../test/drivers/schema_information_test_support')
 const matchers = require('./schema_matchers')
 
+expect.extend({
+    toHaveReadOnlyCapability: matchers.toHaveReadOnlyCapability,
+})
+
```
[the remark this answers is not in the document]
MXPOL: About the boilerplate, we can reduce the boilerplate by adding new property `setupFilesAfterEnv` to the jest config file, in that way the homemade matchers will be available in every test.

The advantage of expanding the expect class is that we can build the matcher array dynamically, compare to 
`matcher+toBe` method.
For example, if  we want to check that every object in the array has the property `x` we need to know the length of the array and then create another array with the same length, and fill the array with the expect object `expect.objectContaining( { x: expect.any() } )`

packages/velo-external-db-core/jest.config.js
```diff
@@ -4,5 +4,6 @@ module.exports = {
     roots: ['<rootDir>/lib'],
     preset: 'ts-jest',
     testRegex: '(.*\\.spec\\.)js$',
-    testEnvironment: 'node'
+    testEnvironment: 'node',
+    setupFilesAfterEnv: ['jest-extended/all'],
```
MXPOL, line 8: Adding new match functions to the expect class

packages/velo-external-db-core/lib/service/schema_matchers.js
```diff
@@ -9,10 +9,20 @@ const appendAllowedOperationsToDbs = (dbs, allowedSchemaOperations) => {
     }))
 }
 
-const toHaveSchemas = ( dbsWithAllowedOperations ) => ({ schemas: dbsWithAllowedOperations.map(asWixSchema) })
+const toHaveSchemas = ( collections, func, ...args ) => ({
```
MXPOL, line 12: `func` is a match function for one collection, it needs to return a matcher (for example we want to test that each collection has a property called `allowedOperations`)

packages/velo-external-db-core/lib/service/schema_matchers.js
```diff
@@ -9,10 +9,20 @@ const appendAllowedOperationsToDbs = (dbs, allowedSchemaOperations) => {
...
+    schemas: collections.map( c => func(c, args) )
+})
 
-const haveSchemaFor = (dbs, allowedSchemaOperations) =>  toHaveSchemas(appendAllowedOperationsToDbs(dbs, allowedSchemaOperations))
+const collectionToHaveReadOnlyCapability = () => expect.objectContaining({ allowedOperations: expect.toIncludeSameMembers(ReadOnlyOperations) })
```
MXPOL, line 16: `toIncludeSameMembers`  is a matcher from the jest-extended package. 
it checks that the arrays have the same members no matter the order

packages/velo-external-db-core/lib/service/schema.spec.js
```diff
@@ -49,7 +50,7 @@ describe('Schema Service', () => {
     test('collections without _id column will have read-only capabilities', async() => {
         driver.givenListResult(ctx.dbsWithoutIdColumn)
 
-        await expect( env.schemaService.list() ).resolves.toEqual( matchers.haveSchemaFor(ctx.dbsWithoutIdColumn, ctx.schemaOperations) )
+        await expect( env.schemaService.list() ).resolves.toEqual( matchers.toHaveReadOnlyCapability(ctx.dbsWithoutIdColumn) )
```
noam-almog, line 53: lets rename to:
```
toHaveReadOnlySchemaLike
```
WDYT ?

PR #143, opened 2022-01-13
Data tests refactor 

MXPOL:
# Reason for This PR

The new feature query validator in `data.js` broke some test, this PR refactors a  little the test to pass the query validator.

### DIFF
--- a/packages/velo-external-db-commons/lib/data_commons.js
+++ b/packages/velo-external-db-commons/lib/data_commons.js
@@ -57,6 +57,19 @@ const AdapterOperators = {
     urlized: 'urlized'
 }
 
+const WixOperators = {
+    $eq: '$eq',
+    $ne: '$ne',
+    $startsWith: '$startsWith',
+    $endsWith: '$endsWith',
+    $contains: '$contains',
+    $hasSome: '$hasSome',
+    $lt: '$lt',
+    $lte: '$lte',
+    $gt: '$gt',
+    $gte: '$gte',
+}
+
 const AdapterFunctions = { 
     avg: 'avg',
     max: 'max',
@@ -105,4 +118,4 @@ const extractProjectionFunctionsObjects = (projection) => projection.filter(f =>
 module.exports = { EMPTY_FILTER, EMPTY_SORT, patchDateTime, asParamArrays, isObject, updateFieldsFor,
                      isEmptyFilter, AdapterOperators, AdapterFunctions,
                      extractGroupByNames, extractProjectionFunctionsObjects,
-                     wixOperatorToAdapterOperator }
+                     wixOperatorToAdapterOperator, WixOperators }
--- a/packages/velo-external-db-commons/lib/data_commons.js
+++ b/packages/velo-external-db-commons/lib/data_commons.js
@@ -65,10 +65,44 @@ const AdapterFunctions = {
     count: 'count'
 }
 
+const wixOperatorToAdapterOperator = operator => {
+    switch (operator) {
+        case '$eq':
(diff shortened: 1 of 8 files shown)


packages/velo-external-db-commons/lib/data_commons.js
```diff
@@ -57,6 +57,19 @@ const AdapterOperators = {
     urlized: 'urlized'
 }
 
+const WixOperators = {
```
MXPOL, line 60: We generate a random Wix operator in `gen.js`, so for the sake of order created a new enum, the random-wix-operator function takes a random value from this enum.

packages/velo-external-db-commons/lib/data_commons.js
```diff
@@ -65,10 +78,44 @@ const AdapterFunctions = {
     count: 'count'
 }
 
+const wixOperatorToAdapterOperator = operator => {
```
MXPOL, line 81: Extract this function from `filter_transformer.js` because I need it also in tests and I didn't want to duplicate myself

packages/velo-external-db-core/lib/service/data.spec.js
```diff
@@ -168,6 +171,8 @@ describe('Data Service', () => {
         ctx.itemId = chance.guid()
         ctx.itemIds = Array.from({ length: 10 }, () => chance.guid())
         ctx.total = chance.natural({ min: 2, max: 20 })
+        ctx.field = gen.randomDbField()
```
MXPOL, line 174: Generate random column, this column will be the only column in the schema that will be return be the mocked 
`CacheableSchemaInformation`

packages/velo-external-db-core/lib/service/data.spec.js
```diff
@@ -168,6 +171,8 @@ describe('Data Service', () => {
...
+        ctx.filterField = gen.randomFilterBaseOnField(ctx.field.field, ctx.field.type)
```
MXPOL, line 175: Because of the  query validator, we need to generate a valid filter (a column that exists in schema, and an operation that can be preformed of the column)

packages/velo-external-db-core/test/drivers/data_provider_test_support.js
```diff
@@ -12,7 +12,7 @@ const dataProvider = {
 }
 
 const givenListResult = (entities, forCollectionName, filter, sort, skip, andLimit) =>
-    when(dataProvider.find).calledWith(forCollectionName, filter, sort, skip, andLimit)
+    when(dataProvider.find).calledWith(forCollectionName, expect.anything(), sort, skip, andLimit)
```
MXPOL, line 15: Here we stub the data-provider find function that `data.js` uses, because of the new filter I need the filter that the `filterTransformer.transform` creates and I don't have it, because of this I changed it to `expect.anything()` 
@noam-almog  Is it terrible?

noam-almog: i don't understand why can't you know what the input is

MXPOL:
Closing the PR, we need a new PR with a different approach: 
 
In `data.js` tests, we don't need to supply the right values for the validate function, we just need to ensure that it's not throwing any error, we can do it by mock the validate function and set the response that the function will return, we are testing `data.js` functions and not the function that it uses ... 

Another thing I think it will be better that `queryValidator.validateFilter` will not throw errors, instead it will just return the fields that doesn't exist or operations that cannot be done on fields, and `data.js` will throw an error depends on the result. (Like I did in `schema.js`)

PR #146, opened 2022-01-17
Mock query validator

Idokah:
Mock `QueryValidator` for data service tests.


packages/velo-external-db-core/test/drivers/filter_transformer_test_support.js
```diff
@@ -20,7 +20,7 @@ const givenFilterByIdWith = (id, filter) => {
                                 })
 }
 
-const stubIgnoreTransform = (filter) => {
+const givenTransformResult = (filter) => {
     when(filterTransformer.transform).calledWith(filter)
                                 .mockReturnValue(filter)
```
noam-almog, line 25: transform needs to return another value, the contract is that filter changes so let's do that here too.

packages/velo-external-db-core/lib/service/data.spec.js
```diff
@@ -125,8 +131,8 @@ describe('Data Service', () => {
 
     test('aggregate api', async() => {
         driver.givenAggregateResult(ctx.entities, ctx.collectionName, ctx.filter, ctx.aggregation)
-        aggregationTransformer.stubIgnoreTransform(ctx.aggregation)
-        filterTransformer.stubIgnoreTransform(ctx.filter)
+        aggregationTransformer.givenAggregateTransformResult(ctx.aggregation)
+        filterTransformer.givenTransformResult(ctx.filter)
```
noam-almog, line 135: naming can be better here, let's think of something that will better reflect what is happening in the test

packages/velo-external-db-core/lib/service/data.spec.js
```diff
@@ -4,24 +4,30 @@ const driver = require('../../test/drivers/data_provider_test_support')
...
+        schema.givenDefaultSchemaFor(ctx.collectionName)
+        filterTransformer.givenTransformResult(ctx.filter)
+
+        const transformedFilter = filterTransformer.transform(ctx.filter)
+        const fields = await schema.schemaInformation.schemaFieldsFor(ctx.collectionName)
+        
+        queryValidator.expectValidFilter(fields, transformedFilter)
 
         const actual = await env.dataService.find(ctx.collectionName, ctx.filter, ctx.sort, ctx.skip, ctx.limit)
         expect( actual ).toEqual({ items: ctx.entities, totalCount: ctx.entities.length })
```
noam-almog, line 25: on the way we need to fix this line to be expect().resolves(....) vs what's written here right now.

packages/velo-external-db-core/lib/service/data.spec.js
```diff
@@ -4,24 +4,30 @@ const driver = require('../../test/drivers/data_provider_test_support')
...
 
 describe('Data Service', () => {
 
-    // eslint-disable-next-line jest/no-disabled-tests
-    test.skip('delegate request to data provider and translate data to velo format', async() => {
+    test('delegate request to data provider and translate data to velo format', async() => {
         driver.givenListResult(ctx.entities, ctx.collectionName, ctx.filter, ctx.sort, ctx.skip, ctx.limit)
-        filterTransformer.stubIgnoreTransform(ctx.filter)
+        schema.givenDefaultSchemaFor(ctx.collectionName)
+        filterTransformer.givenTransformResult(ctx.filter)
```
noam-almog, line 17:
```
givenTransformTo( ctx.domainFilter, ctx.filter)
```

PR #147, opened 2022-01-18
Validate dbtype and cloud vendor

### LINKED ISSUE #119: Notify when cloud vendor / db type is wrong
When filling wrong `cloud_vendor` there's no appropriate message or warning : 
image

When filling wrong db type the adapter is collapsing.

Idokah:
when filling wrong cloud vendor the message was "External DB config read successfully" 
when filling wrong dbType the adapter was collapsing.
(you can see in the linked issue)

I added in the `CommonConfigReader.validate` another validation - validType, validVendor. 

And I made stub-db with the required classes like every other implementation to allow the adapter to load successfully. 
I'm not sure where it belongs, for now it's in `packages/velo-external-db/lib/storage/stub-db` but i know it's not the right place for it.
do you have an idea where it should be? maybe i should expend it and make it a package like any other implementation? 

image

fix #119

### DIFF
--- a/packages/external-db-config/lib/utils/config_utils.js
+++ b/packages/external-db-config/lib/utils/config_utils.js
@@ -2,4 +2,8 @@ const objectContainsKey = (obj, key) => typeof obj[key] === 'string' && obj[key]
 
 const checkRequiredKeys = (obj, requiredKeys) => requiredKeys.filter(key => !objectContainsKey(obj, key) )
 
-module.exports = { checkRequiredKeys }
+const supportedDBs = ['postgres', 'spanner', 'firestore', 'mssql', 'mysql', 'mongo', 'airtable', 'dynamodb', 'bigquery']
+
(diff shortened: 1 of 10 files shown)


packages/external-db-config/lib/utils/config_utils.js
```diff
@@ -2,4 +2,8 @@ const objectContainsKey = (obj, key) => typeof obj[key] === 'string' && obj[key]
 
 const checkRequiredKeys = (obj, requiredKeys) => requiredKeys.filter(key => !objectContainsKey(obj, key) )
 
-module.exports = { checkRequiredKeys }
+const supportedDBs = ['postgres', 'spanner', 'firestore', 'mssql', 'mysql', 'mongo', 'airtable', 'dynamodb', 'bigquery']
```
MXPOL, line 5: google-sheet ?

PR #155, opened 2022-01-23
config reader for auth configuration

MXPOL:


packages/external-db-config/lib/readers/common_config_reader.js
```diff
@@ -4,8 +4,8 @@ class CommonConfigReader {
     constructor() { }
 
     readConfig() {
-        const { CLOUD_VENDOR, TYPE, REGION, SECRET_NAME } = process.env
-        return { vendor: CLOUD_VENDOR, type: TYPE, region: REGION, secretId: SECRET_NAME }
+        const { CLOUD_VENDOR, AUTH_VENDOR, TYPE, REGION, SECRET_NAME } = process.env
+        return { vendor: CLOUD_VENDOR, authVendor: AUTH_VENDOR, type: TYPE, region: REGION, secretId: SECRET_NAME }
```
MXPOL, line 8: @noam-almog  not sure if we need to verify `AUTH_VENDOR` variable (check that it was supplied) 
Will there always be authorization?
WDYT?

noam-almog: auth vendor is the same as vendor, no ?

MXPOL: We agreed that these would be 2 different things …

packages/external-db-config/lib/readers/aws_auth_config_reader.js
```diff
@@ -0,0 +1,36 @@
+const { SecretsManagerClient, GetSecretValueCommand } = require('@aws-sdk/client-secrets-manager')
+const { checkRequiredKeys } = require('../utils/config_utils')
+
+const EmptyAWSAuthConfig = { callbackurl: '', clientid: '', clientsecret: '', clientdomain: '' }
+
+class AwsAuthConfigReader {
```
noam-almog, line 6: do you want to define a different secret key for auth ? 
i want to use the same key as before but auth will be a sub-property in the current config.

packages/external-db-config/lib/readers/common_config_reader.js
noam-almog: why ? it's the same value ? aws/gcp/azure, no ?

packages/external-db-config/lib/readers/aws_auth_config_reader.js
MXPOL: I can't set the value for `clientsecret`, it's a secret that generated when I create the login screen in AWS/GCP

MXPOL: image

Issue #150, opened 2022-01-18
Bug when opening collections without _id column in the content manger 


MXPOL:
When I open a collection without `_id` column I do not get the rows of the collection but only the first row duplicated endlessly
As you can see in the picture below

![image](https://user-images.githubusercontent.com/407079/149983832-2c7af37d-edab-4a74-b083-f5e6b4ccf91e.png)

MXPOL:
#151  fixes this bug, by generating random _id property to objects without _id. 

In the next PR, we're planning to use the row number or the key primary of the collection as the _id property

PR #156, opened 2022-01-25
hash id

Idokah:
Hash the _id with sha1-base64 of concatenation of all the item fields - to get more consistence _id.

### DIFF
--- a/packages/velo-external-db-core/lib/converters/transform.js
+++ b/packages/velo-external-db-core/lib/converters/transform.js
@@ -25,7 +25,7 @@ const generateIdsIfNeeded = item => {
     if ('_id' in item)
         return item
     const sha = crypto.createHash('sha1')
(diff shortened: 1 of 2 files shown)


packages/velo-external-db-core/lib/converters/transform.js
```diff
@@ -21,10 +22,11 @@ const unpackDates = item => {
 }
 
 const generateIdsIfNeeded = item => {
-    if ('_id' in item) {
+    if ('_id' in item)
         return item
-    }
-    return { ...item, _id: uuidv4() }
+    const sha = crypto.createHash('sha1')
+    const fieldsConcat = Object.keys(item).reduce((pV, key) => (pV.concat(item[key])), '')
```
MXPOL, line 28: There is a simpler way : `Object.values(item).join('')`

Idokah:
fix #150

PR #157, opened 2022-01-25
Schema Aware Data Service

Idokah:
Creating `PreDataService`,  It's schema-aware layer that transform and validate filters and aggregations (didn't implement the aggregation validation yet.)
And prepare the items, for update/insert, for `DataService`.


packages/velo-external-db-core/lib/service/pre_data.js
```diff
@@ -0,0 +1,42 @@
+const { prepareForUpdate, unpackDates, prepareForInsert } = require('../converters/transform')
+
+class PreDataService {
```
noam-almog, line 3: lets call this SchemaAwareDataService

noam-almog:
i don't want router to contain all that logic, 
let's go with a different approach, router will call SchemaAwareDataService (find) and it will call find on data service with all of the objects after transformation (and transform the response if needed)
adding logic to router it not my favourite idea.

Idokah:
> i don't want router to contain all that logic, let's go with a different approach, router will call SchemaAwareDataService (find) and it will call find on data service with all of the objects after transformation (and transform the response if needed) adding logic to router it not my favourite idea.

I think It will create another big class with a lot of logic, like the old data service, the only difference be that he call data service instead data provider, and if the class will  transform the response as well there is no point to have data service, we can just pass data provider, and we'll get the same class as before.

I have another suggestion, It's a bit similar to what I did, but what do you think about export this logic to middleware - the middleware will recognize `filter`, `aggregate` in the request body,  transform and validate them and change the request  body before it reaches the call to `dataService` from router.

And the same with `prepareForInsert`, `prepareForUpdate`.

noam-almog:
the logic is very defined, it tells us that it's the same service but it normalized the data with the current schema, what you created is more in the area of a util class which i think will work less for us in the long run, lets discuss the options tomorrow.

Idokah:
I changed the class to the way we agreed on. 
I still need to implement validate aggregation

PR #161, opened 2022-01-26
Integration of the auth reader in config reader

MXPOL:
This PR integrates the auth reader in the config reader instead of 2 config readers. 
**Breaking change**: `configReader.configStatus()` returns an object with 3 properties : validAuthConfig, validConfig, message instead of just a string..


packages/external-db-config/lib/service/config_reader.js
```diff
@@ -1,31 +1,44 @@
...
     
-    else if (!validVendor) {
-      return 'Cloud type is not supported'
-    }
-
-    else if(!validType) {
-      return 'DB type is not supported'
-    }
+  
+    let message
```
MXPOL, line 24: The login screen and app info will use `configStatus` to print the message if something is missing,
Hence, the order of the ifs ....

Issue #159, opened 2022-01-26
Add projection to domain filter object


Idokah:
In case of `data/find`
If we'll not get a projection (and we will not for now) we'll send projection with all the fields 

today:
```
filter:
{
  fieldName: 'theFieldName'
  operator: adapterOperator
  value: 'val'
}
```
with projection: 
```
filter:
{
  projection: [ { name: '_id' } , { name: '_createdDate'} , ... ]
  fieldName: 'theFieldName'
  operator: adapterOperator
  value: 'val'
}
```

I decided the projection would be an array of objects because I want to allow the option of aliases similarly to aggregation.
```
projection: [{name: 'field1', alias: 'someAlias'}]
```

In case of `data/aggregation` we already have a projection : 
```
{
  filter: { }, 
  projection: [{name:'groupByFieldName'}, 
                {name:'groupByFieldName2'}, 
                {name: 'field1', alias:'field1Max', function: 'max'} ]
  postFilter: { }
}
```

noam-almog:
in normal case, is projection part of the filter ? if so, i don't think they are related, can we split them ?

Idokah:
> in normal case, is projection part of the filter ? if so, i don't think they are related, can we split them ?

You're right, I split it.
And I changed the projection to be an array of fields without aliases like we talked.

PR #163, opened 2022-01-30
Projection - Postgres

Idokah:
Related to #159 
Add the option to add to the request of `data/find`, `data/get` projection parameter.
If the `SchemaAwareData` we are calling  `createProjectionArr` (not sure about the name) that do this according to the projection array input:

- if the projection array is valid (all the fields are really exists) it's returning the same array.
- if the projection array includes other fields beside what's really exists, it's returning the same array without the non-existed fields.
- if the projection array is empty / the user has not set this parameter at all or any other bad input, it's returning an array of all the fields of the schema.

I've added tests in `data_provider.spec.js`, Postgres - `sql_filter_transformer.spec.js`. 

LMK what you think.

### DIFF
--- a/packages/external-db-postgres/lib/sql_filter_transformer.js
+++ b/packages/external-db-postgres/lib/sql_filter_transformer.js
@@ -245,6 +245,11 @@ class FilterParser {
         }]
     }
 
+    projection(projection) {
+        if (!projection) return '*'
+        return projection.map(escapeIdentifier).join(', ')
+    }
+
     patchTrueFalseValue(value) {
         if (value === true || value === false) {
             return value ? 1 : 0
--- a/packages/external-db-postgres/lib/sql_filter_transformer.js
+++ b/packages/external-db-postgres/lib/sql_filter_transformer.js
@@ -245,8 +245,7 @@ class FilterParser {
         }]
     }
 
-    projection(projection) {
-        if (!projection) return '*'
+    selectFieldsFor(projection) {
         return projection.map(escapeIdentifier).join(', ')
     }
 
--- a/packages/velo-external-db-core/lib/converters/transform.js
+++ b/packages/velo-external-db-core/lib/converters/transform.js
@@ -1,6 +1,7 @@
 const { v4: uuidv4 } = require('uuid')
 let dateTimeProvider = require('../utils/date_time_provider')
 const crypto = require('crypto')
+const { isEmptyProjection } = require('velo-external-db-commons')
 
 const asWixData = e => generateIdsIfNeeded(packDates(e))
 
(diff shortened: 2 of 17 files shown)


packages/external-db-postgres/lib/sql_filter_transformer.js
```diff
@@ -245,6 +245,11 @@ class FilterParser {
         }]
     }
 
+    projection(projection) {
+        if (!projection) return '*'
```
Idokah, line 249: I'm making sure that the projection will not be empty when it's arriving here (in `schema_aware` layer)
I should remove this line?

packages/external-db-postgres/lib/sql_filter_transformer.js
```diff
@@ -245,6 +245,11 @@ class FilterParser {
         }]
     }
 
+    projection(projection) {
```
noam-almog, line 248: a better name would be selectFieldsFor

packages/external-db-postgres/lib/sql_filter_transformer.js
noam-almog, line 249: do we get to this case ? we will always send projection so why add this if ?

packages/velo-external-db-core/lib/converters/transform.js
```diff
@@ -54,6 +55,22 @@ const defaultValueFor = (f) => {
 const prepareForInsert = (item, fields) => fields.reduce((pv, f) => ({ ...pv, [f.field]: item[f.field] || defaultValueFor(f) }), {})
 const prepareForUpdate = (item, fields) => fields.reduce((pv, f) => f.field in item ? ({ ...pv, [f.field]: item[f.field] }) : pv, {})
 
+
+const createProjectionArr = (fields, projection) => {
+    const fieldsArr = fields.map(f => f.field)
+ 
+    if (isEmptyProjection(projection)) return fieldsArr
```
noam-almog, line 62: can we have an empty array here ? why ? is there a case where a table doesn't have any fields ?

packages/velo-external-db-core/lib/service/schema_aware_data.js
```diff
@@ -7,13 +7,16 @@ class SchemaAwareDataService {
         this.dataService = dataService
     }
 
-    async find(collectionName, filter, sort, skip, limit) {
+    async find(collectionName, filter, sort, skip, limit, _projection) {
```
noam-almog, line 10: currently we don't send projection from router, so why add this param here ?
we didn't change the api, so currently the projection is just an internal thing, please remove this.

packages/velo-external-db-core/lib/service/schema_aware_data.js
```diff
@@ -85,6 +87,11 @@ class SchemaAwareDataService {
                     .map(i => unpackDates(i))
     }
 
+    async createProjectionArr(collectionName, projection) {
```
noam-almog, line 90: function name can be better, i suggest schemaFieldNamesFor

noam-almog:
i want to reduce complexity of this PR, let's assume the api integration will happen in the future, for now, we want to have schema aware automatically add the schema fields to the flow and have the provider use it.
we will do a second PR of how we want to currently integrate this into the API

packages/velo-external-db-core/lib/service/schema_aware_data.js
Idokah: I think my function name isn't good as well, but your suggestion is good for now because it's internal, but when we'll get the projection from outside it will not choose all the schema field names for the collection. 
the purpose of the function is to clean up the projection it receives from non-existent fields, and if the projection is not valid - return the default (all the schema fields).

Idokah: Do you want me to treat it as internal only  for now, and remove this logic?

packages/velo-external-db-core/lib/converters/transform.js
Idokah: Same here as below, I treated the projection as input from outside, and not as all the fields

packages/velo-external-db-core/lib/service/schema_aware_data.js
noam-almog: in our case, if the fields would be invalid we will reject the request and not try to fix it.

packages/velo-external-db-core/lib/converters/transform.js
noam-almog: so right now the logic is, get al fields and pass it to the next layer, in the future we will handle the customer input.
in case the customer doesn't add any fields, we will need to decide how to behave, i will prefer to reject the request rather than trying to execute it.

PR #164, opened 2022-01-31
Validate before getById 

Idokah:
Add validation before `getById`, export `getByIdFilter` to utils.

### DIFF
--- a/packages/velo-external-db-core/lib/converters/query_validator.js
+++ b/packages/velo-external-db-core/lib/converters/query_validator.js
@@ -1,5 +1,6 @@
 const { InvalidQuery } = require('velo-external-db-commons').errors
 const { extractFieldsAndOperators, queryAdapterOperatorsFor } = require ('./query_validator_utils')
+const { getByIdFilterFor } = require ('../utils/data_utils')
 
 class QueryValidator {
     constructor() {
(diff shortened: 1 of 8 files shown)


packages/velo-external-db-core/lib/service/schema_aware_data.js
```diff
@@ -13,6 +14,7 @@ class SchemaAwareDataService {
     }
 
     async getById(collectionName, itemId) {
+        await this.validateFilter(collectionName, getByIdFilter(itemId))
```
noam-almog, line 17: add test for this and also validate cases of invalid id, empty string/null value should throw an exception

packages/velo-external-db-core/lib/converters/query_validator.spec.js
```diff
@@ -51,6 +51,24 @@ describe('Query Validator', () => {
         })
     })
 
+    describe ('validateGetById', () => {
+        test('should throw InvalidQuery if itemId is not defined', () => {
+            expect ( () => env.queryValidator.validateGetById(ctx.fieldArrWithId)).toThrow(InvalidQuery)
+            expect ( () => env.queryValidator.validateGetById(ctx.fieldArrWithId, '')).toThrow(InvalidQuery)
```
noam-almog, line 57: also add a case '    '

packages/velo-external-db-core/lib/converters/query_validator.js
```diff
@@ -15,6 +16,11 @@ class QueryValidator {
     
     }
     
+    validateGetById(fields, itemId) {
+        if (!itemId) throw new InvalidQuery('A value must be provided for itemId')
+        this.validateFilter(fields, getByIdFilterFor(itemId))
```
noam-almog, line 21: this logic seems funny, we are validating a filter we synthetically create, i don't see the value in this. WDYT ?

PR #171, opened 2022-02-03
AuthenticationService class 

MXPOL:
Created Authentication class in services folder, this class responsible for all things related to authentication including auth routes (login, callback, logout) and also authentication middleware that checks if the user is logged in and if no it will redirect the user to the login screen.

Pros of this change: `velo-external-db` module doesn't have any package that related to authentication


packages/external-db-authorization/lib/factory.js
```diff
@@ -1,37 +1,33 @@
...
 }
 
-const createAuthService = async(vendor, config) => {
+const initAuthProvider = async(vendor, config) => {
 
   const { auth: authConfig } = await config.readConfig()
-  const { validAuthConfig } = await config.configStatus()
+  const authInformation = await config.configStatus()
+
+  const authProvider = authProviderFor(vendor, authConfig)
```
Idokah, line 27: If the case was default (local/azure) , so you receive an object back from that function, how do you handle this response? I don't find it in the code

MXPOL:
@noam-almog  ready for review, can you check it?

packages/external-db-authorization/lib/services/authentication_service.js
```diff
@@ -0,0 +1,73 @@
...
+        return this.authRouter
+    }
+
+    createLoginHandler() {
+
+        this.authRouter.use(this.loginPath, passport.authenticate(this.authName))
+    }
+
+    createLoginCallbackHandler() {
+        const redirectOptions = { successReturnToOrRedirect: '/', failureRedirect: '/auth/login' }
```
Idokah, line 42: I'm not sure that it supposed to be in the constructor (maybe for the tests you tell me), 
but have `this.loginPath` in constructor, maybe use it here in `failureRedirect` instead hard-coded?

packages/external-db-authorization/package.json
```diff
@@ -25,11 +25,13 @@
   },
   "dependencies": {
     "@aws-sdk/client-cognito-identity-provider": "^3.47.0",
+    "express": "^4.17.2",
```
noam-almog, line 28: we should not import express here, i thought the security model should include passport here, but not express

MXPOL: We decided that the authentication will be middleware, therefore I have to import Express and use it in the packge ..

noam-almog: middleware is related to express therefor should be on the main module, the api's being called in it should be on the authorization module
(and change the module to security on another PR)

PR #172, opened 2022-02-06
New validation function for gcp auth

MXPOL:
This PR adds validation for GCP authentication mechanism, after the user redirects from Google back to the app, the function ensure that the user is part of the project owner, if so he will pass to the app info screen.
If not, the user will be thrown out and the identification process will not be completed

**breaking change -** new role (Security Reviewer) must be assigned to the cloud run service account, otherwise the authentication in GCP will not work

### DIFF
--- a/packages/external-db-authorization/lib/auth-providers/gcp_auth_provider.js
+++ b/packages/external-db-authorization/lib/auth-providers/gcp_auth_provider.js
@@ -10,31 +10,31 @@ class GcpAuthProvider {
       scope: ['email', 'profile']
     }
 
-    return new GoogleStrategy(this.options, this.verify)
+    return new GoogleStrategy(this.options, this.verify.bind(this))
   }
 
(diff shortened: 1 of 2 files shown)


packages/external-db-authorization/lib/strategies/gcp_strategy.js
```diff
@@ -12,8 +12,27 @@ class GcpStrategy {
     return new GoogleStrategy(this.options, this.verify)
   }
 
-  verify(AccessToken, tokenSecret, profile, done) {
-        done(null, profile)
+  verify(accessToken, tokenSecret, profile, cb) {
+    const usersEmail = profile._json.email
+
+    const getProjectOwners = async() => {
```
MXPOL, line 18: Couldn't move this function to `gcpStrategy` class,  verify function is being used by passport package, and If i will move this function to the class I will have to use `this` keyword that will be **different** when passport will run this function ...

PR #178, opened 2022-02-09
Collection level role config

Idokah:
Get the collection role config from env (need to set them through the secret-manger / app config) 

example for the json that we are expecting in `process.env.ROLE_CONFIG`: 

```
collectionLevelConfig : [
  {
    "id":"collectionName1"
    "readPolicies": ["OWNER","BACKEND_CODE"],
    "writePolicies": ["OWNER","BACKEND_CODE"]
  },
  {
    "id":"collectionName2"
    "readPolicies": ["OWNER","BACKEND_CODE", "VISITOR"],
    "writePolicies": ["OWNER","BACKEND_CODE", "VISITOR"]
  }
]

```

### DIFF
--- a/packages/external-db-config/lib/readers/authorization_config_reader.js
+++ b/packages/external-db-config/lib/readers/authorization_config_reader.js
@@ -1,0 +1,20 @@
+const { checkRequiredKeys, isJson, EMPTY_ROLE_CONFIG } = require('../utils/config_utils')
+
+class AuthorizationConfigReader {
+  constructor() {
+  }
+
+  async readConfig() {
+    const { ROLE_CONFIG } = process.env
+    const { collectionLevelConfig } = isJson(ROLE_CONFIG) ? JSON.parse(ROLE_CONFIG) : EMPTY_ROLE_CONFIG
+    return collectionLevelConfig
+  }
+
+  validate() {
+    return {
+      missingRequiredSecretsKeys: checkRequiredKeys(process.env, [])
+    }
+  }
+}
+
+module.exports = AuthorizationConfigReader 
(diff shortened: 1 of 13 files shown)


packages/external-db-config/lib/readers/authorization_config_reader.js
```diff
@@ -0,0 +1,20 @@
+const { checkRequiredKeys, isJson, EMPTY_ROLE_CONFIG } = require('../utils/config_utils')
+
+class AuthorizationConfigReader {
+  constructor() {
+  }
+
+  async readConfig() {
+    const { ROLE_CONFIG } = process.env
+    const { collectionLevelConfig } = isJson(ROLE_CONFIG) ? JSON.parse(ROLE_CONFIG) : EMPTY_ROLE_CONFIG
```
noam-almog, line 9: what happens in case the json is not parsing or contain bad data ?

noam-almog:
do we plan to include app info representation of the auth data in this PR too ?

packages/external-db-config/lib/readers/authorization_config_reader.js
Idokah: Empty role config.
Default permissions for all the collections

noam-almog: user can provide invalid json and not be aware of that.
what i mean is a valid json, but an invalid data.
we need to validate it as well and inform the user of that.

Idokah:
> do we plan to include app info representation of the auth data in this PR too ?

I think I'll do it in another PR. I don't want this PR to be bigger.

packages/external-db-config/lib/readers/authorization_config_reader.js
Idokah: WDYT that we need to validate for example? 
that the JSON is in the right format (`id`,` readPolicies`,` writePolicies`) ?

noam-almog:
yes, validate the format

PR #187, opened 2022-02-15
Fix MongoDB add / remove column concurrency

Idokah:


noam-almog:
revert the promise all change back on this pr

Idokah:
I think the issue is still exists in Firestore, so maybe we'll fix this first.

Idokah:
@noam-almog 
And besides, there's no way to call `addColumn` a few times without await (correct me if I'm wrong), so I think other than that it looks a little better, it does not simulate a possible behavior of the adapter and can cause bugs that unique only to the script. 
maybe there is better solution than `promise.All`

PR #196, opened 2022-02-17
patch generate _id

### LINKED ISSUE #192: _id is not generated if omitted in insert and bulkInsert API
This code was working previously and now fails if _id is not explicitly set.

`import wixData from 'wix-data';

export async function insertTest(event) {

	// const item = {column_0: 'insert test1', column_1: 'test 1', _id: 'kkk1'} ; // works
	const item = {column_0: 'insert test1', column_1: 'test 1'} ; // fails with unknown error

	const res = await wixData.insert('velo-demo/demo1', item);
`

Idokah:
fix #192 
We used `isPrimary` property to decide whether generate uuid or not.
We expect to receive this `isPrimary` property from `schemaProvider` but for now there is no implementation that returns that prop.

### DIFF
--- a/packages/velo-external-db-core/lib/converters/item_transformer.js
+++ b/packages/velo-external-db-core/lib/converters/item_transformer.js
@@ -37,7 +37,7 @@ class ItemTransformer {
                 }
                 return 0  // default for number is 0 
             case 'text':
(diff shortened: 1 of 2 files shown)


noam-almog:
please add it to item transformer unit test

PR #201, opened 2022-02-22
rename const

noam-almog:


Idokah:
The convention for docker environment variables is ROLE_CONFIG, you can see this in our project - all the envs are with capital letters,
and you can even see it in our docker-compose [file](https://github.com/wix/velo-external-db/blob/master/packages/velo-external-db/test/resources/docker-compose.yaml#L62), all the envs are in that convention. 
I don't think that the user should notice that our project is in JS and because of that we are using camelCase for envs as well, it's just a docker for him and that one of the envs is in camelCase is just weird. 
At least all the variables were in this convention, but when it's the only one, I do not think it's a good decision.

Idokah:
But of course it's not a matter of life, so if you still think it's the right way, you can merge this.

PR #202, opened 2022-02-22
removed each from e2e and storage tests

michaelir:


packages/velo-external-db/test/e2e/app.e2e.spec.js
```diff
@@ -1,30 +1,29 @@
 const { authOwner } = require('../drivers/auth_test_support')
-const each = require('jest-each').default
-const { initApp, teardownApp, dbTeardown, testSuits } = require('../resources/e2e_resources')
+const { initApp, teardownApp, dbTeardown } = require('../resources/e2e_resources')
+const { name, setup } = require('../resources/e2e_resources').testedSuit()
```
noam-almog, line 3: why not just call the code inside of the describe, instead of here.
on the place that you will need the setup, just call it again.

packages/velo-external-db/test/resources/e2e_resources.js
```diff
@@ -55,23 +55,32 @@ const firestoreTestEnvInit = async() => await dbInit(firestore)
...
-    ['Firestore', firestoreTestEnvInit],
-    ['Sql Server', mssqlTestEnvInit],
-    ['Mongo', mongoTestEnvInit],
-    ['Google-sheet', googleSheetTestEnvInit],
-    ['Airtable', airtableTestEnvInit],
-    ['DynamoDb', dynamoTestEnvInit],
-    ['BigQuery', bigqueryTestEnvInit]
-].filter( ([name]) => name.toLowerCase() === process.env.TEST_ENGINE || (name === 'Sql Server' && process.env.TEST_ENGINE === 'mssql') )
+const testSuits = {
+    mysql: ['MySql', mysqlTestEnvInit],
```
noam-almog, line 64: lets convert the array to an object, { name: 'MySql', target: 'mysql',  setup: mysqlTestEnvInit }
this way we can have more clarity on which is what are we using and comparing, WDYT ?

michaelir: Agree

packages/velo-external-db/test/e2e/app.e2e.spec.js
```diff
@@ -1,30 +1,28 @@
...
 
 const axios = require('axios').create({
     baseURL: 'http://localhost:8080'
 })
 
-describe('Velo External DB',  () => {
-    each(testSuits()).describe('%s', (name, setup) => {
-        beforeAll(async() => {
-            await setup()
+describe(`Velo External DB: ${testedSuit().name}`,  () => {
```
noam-almog, line 8: i would switch it to 
``` ${currentDbImplementationName()}```

File: packages/velo-external-db/test/e2e/app.e2e.spec.js [hunk 4]
```diff
@@ -1,30 +1,28 @@
...
+    beforeAll(async() => {
+        await testedSuit().setup()
```
noam-almog, line 10: ``` setupDb() ```

packages/test-commons/lib/test-commons.js
```diff
@@ -1,6 +1,8 @@
 const { SchemaOperations } = require('velo-external-db-commons')
 const { AddColumn, RemoveColumn } = SchemaOperations
 
+const suitDef = (name, setup) => ( { name, setup } )
```
noam-almog, line 4: not sure why this is needed here, why not just inline it ? or just write the config on the resources js at you want it ?

PR #207, opened 2022-02-27
Terraform scripts - GCP

MXPOL:
Terraform script that creates an adapter in GCP

### DIFF
--- a/scripts/terraform/gcp/variables.tf
+++ b/scripts/terraform/gcp/variables.tf
@@ -4,10 +4,16 @@ variable "project_id" {
   type        = string
 }
 
-variable "location" {
+variable "adapter_location" {
   description = "Velo External DB adapter service deployment location"
   type        = string
-  default = "us-central1"
+  default = "us-east4"
+}
+
+variable "db_location" {
+  description = "Database location"
+  type        = string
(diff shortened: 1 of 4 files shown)


scripts/terraform/gcp/variables.tf
```diff
@@ -0,0 +1,35 @@
+
+variable "project_id" {
+  description = "The project ID to deploy to"
+  type        = string
+}
+
+variable "location" {
+  description = "Velo External DB adapter service deployment location"
+  type        = string
+  default = "us-central1"
```
Idokah, line 10: us-east4!!!

scripts/terraform/gcp/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+
+variable "image_url" {
+  description = "GCR hosted image URL to deploy"
+  type        = string
+  default = "gcr.io/wix-velo-api/velo-external-db"
+}
+
+variable "secret_key" {
+  description = "Secret Key for Velo External DB adapter service"
+  type        = string
```
Idokah, line 27: Maybe add here 
`sensitive   = true`
and remove default

PR #212, opened 2022-03-01
Case insensitive tests

Idokah:
Change the `startsWith`/ `endsWith`/ `contains` to be case-insensitive 
 I changed it for all implementations except the following:
**Firestore**, **Google-Sheets**: @MXPOL needs to implement, and he will do this in another PR.
**Airtable**:  I'll do it in another PR, I'm facing some problems there.
**DynamoDB** - don't support case-insensitive, you can see in this [thread](https://forums.aws.amazon.com/thread.jspa?threadID=92159), so for now I disabled the case-insensitive test for it.

### DIFF
--- a/packages/velo-external-db/test/drivers/data_provider_matchers.js
+++ b/packages/velo-external-db/test/drivers/data_provider_matchers.js
@@ -1,7 +1,3 @@
 const entitiesWithOwnerFieldOnly = (entities) => expect.arrayContaining(entities.map(e => ({ _owner: e._owner })))
 
-
-const toggleCase = (str) => str.split('').map(c => c === c.toUpperCase() ? c.toLowerCase() : c.toUpperCase()).join('')
-
-module.exports = { entitiesWithOwnerFieldOnly, toggleCase }
-
+module.exports = { entitiesWithOwnerFieldOnly }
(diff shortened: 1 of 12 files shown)


packages/velo-external-db/test/storage/data_provider.spec.js
```diff
@@ -53,6 +53,16 @@ describe(`Data API: ${currentDbImplementationName()}`, () => {
         })
     }
 
+    if (shouldRunOnlyOn(['Postgres', 'MySql'], currentDbImplementationName())) {
```
noam-almog, line 56: add one test for case sensitive and one for case insensitive let's cover all the bases.

packages/velo-external-db/test/drivers/data_provider_matchers.js
```diff
@@ -1,4 +1,7 @@
 const entitiesWithOwnerFieldOnly = (entities) => expect.arrayContaining(entities.map(e => ({ _owner: e._owner })))
 
-module.exports = { entitiesWithOwnerFieldOnly }
+
+const toggleCase = (str) => str.split('').map(c => c === c.toUpperCase() ? c.toLowerCase() : c.toUpperCase()).join('')
```
noam-almog, line 4: why do you need string split here ? it should work on a string with spaces.
plus, let's inline this, no need to add a function for that.

Idokah: I did it to change string like 'Yo' to 'yO', but it's not necessary because I'm toggling only the first char. 

I put it into the test instead of the function, but not as inline on the calling to `givenStartsWithFilterFor` because it’s seemed to me not clear enough. WDYT ?

PR #214, opened 2022-03-01
Retrieve schema operations in tests

MXPOL:


packages/velo-external-db/test/drivers/schema_api_rest_test_support.js
```diff
@@ -9,6 +11,14 @@ const givenCollection = async(name, columns, auth) => {
     }
 }
 
+const retrieveSchemaOperations = async() => {
+    const tempCollectionName = 'tempCollection'
+    await axios.post('/schemas/create', { collectionName: tempCollectionName }, authOwner)
+    const  { schemas }   = (await axios.post('/schemas/find', { schemaIds: [ tempCollectionName ] }, authOwner)).data
```
Idokah, line 17: Maybe instead these lines because there are already functions that does that : 

```
await givenCollection('tempCollection',authOwner)
const {schemas} = await retrieveSchemaFor('tempCollection', authOwner).data
...

```

Idokah:
`    test('find api', async() => testIfSchemaSupportsFindWithSort(env, async() => {`

I think it will be better if you pass to the function only what is needed, instead of `env` pass `env.schemaOperations`.
It will be easier to find the use of this variable in the code, and it will be clearer when you look at the test what the function does exactly.

packages/velo-external-db/test/drivers/schema_api_rest_test_support.js
```diff
@@ -9,6 +11,14 @@ const givenCollection = async(name, columns, auth) => {
     }
 }
 
+const retrieveSchemaOperations = async() => {
+    const tempCollectionName = 'tempCollection'
+    await axios.post('/schemas/create', { collectionName: tempCollectionName }, authOwner)
```
noam-almog, line 16: why do we need to create a collection to retrieve schema capabilities, don't we have that even if we don't have any collections ?

MXPOL:
Closing this PR, 
 I went for a different approach, so I opened new [PR](https://github.com/wix/velo-external-db/pull/235) for this.

PR #218, opened 2022-03-02
Working startWith operator - Firestore

MXPOL:
Turns out that the `startWith` operator did not work properly, This PR fixed it

### DIFF
--- a/packages/external-db-firestore/lib/firestore_utils.js
+++ b/packages/external-db-firestore/lib/firestore_utils.js
@@ -1,5 +1,6 @@
 const { Timestamp } = require('@google-cloud/firestore')
 
+const lastLetterCode = String.fromCharCode(65535)
 
 const fixDates = (value) => {
     if (value instanceof Timestamp) {
@@ -16,4 +17,6 @@ const asEntity = (docEntity) => {
   }.bind(this), {})
 }
 
-module.exports = { fixDates, asEntity }
+
+
+module.exports = { fixDates, asEntity, lastLetterCode }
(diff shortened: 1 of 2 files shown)


packages/external-db-firestore/lib/sql_filter_transformer.js
```diff
@@ -130,6 +134,19 @@ class FilterParser {
...
+    parseStringBegins(fieldName, inlineFields, value) {
+        return [{
+            fieldName: this.inlineVariableIfNeeded(fieldName, inlineFields),
+            opStr: '>=',
+            value,
+        },
+        {
+            fieldName: this.inlineVariableIfNeeded(fieldName, inlineFields),
+            opStr: '<',
+            value: value + String.fromCharCode(65535)
```
Idokah, line 146: `String.fromCharCode(65535)` should be at firestore_utils as `MaxChar` in my opinion

noam-almog: pack this value to a function or const with proper name so we can understand what you did there.

packages/external-db-firestore/lib/firestore_utils.js
```diff
@@ -1,5 +1,6 @@
 const { Timestamp } = require('@google-cloud/firestore')
 
+const lastLetterCode = String.fromCharCode(65535)
```
Idokah, line 3: That's constant - I would rename it to `LastLetterCoder`

packages/external-db-firestore/lib/sql_filter_transformer.js
```diff
@@ -49,9 +54,13 @@ class FilterParser {
     isSingleFieldOperator(operator) {
         return [ne, lt, lte, gt, gte, include, eq, string_begins, string_ends].includes(operator)
     }
+
+    isStringBeginsOperator(operator) {
+        return operator === string_begins
+    }
```
Idokah, line 60: I think you can check this condition in the if statement.

MXPOL: I prefer it to be clear what we are actually checking

PR #223, opened 2022-03-09
Fix dynamo config reader

Idokah:


packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -48,15 +49,18 @@ class AwsConfigReader {
         return { region: this.region, secretKey: process.env.SECRET_KEY, endpoint: process.env.ENDPOINT_URL }
       }
       const cfg = await this.readExternalConfig()
-                            .catch(() => EmptyAWSConfig)
+                            .catch(() => { true})
```
MXPOL, line 52: what is `{ true }`?

Idokah: typo, I meant `{ error:true }`, it's the empty config of dynamo.

PR #231, opened 2022-03-15
Change aws config reader

### LINKED ISSUE #226: Allow insert environment variables in AWS not only through secret manager
All clouds allow users to choose whether they want to use the secret manger or insert the environment variables through the app configuration, except AWS.

Idokah:
fix #226 
Add the ability to read hybrid configuration in AWS - part of it from environment variables, and part of it from the secret manager.

### DIFF
--- a/packages/external-db-config/lib/readers/aws_config_reader.js
+++ b/packages/external-db-config/lib/readers/aws_config_reader.js
@@ -42,7 +42,9 @@ class AwsDynamoConfigReader {
 
     async readConfig() {
       const { config } = await this.getExternalAndLocalEnvs()
-  
+      if (process.env.NODE_ENV === 'test') {
+        return { region: this.region, secretKey: config.SECRET_KEY, endpoint: process.env.ENDPOINT_URL }
+      }
       return { region: this.region, secretKey: config.SECRET_KEY }
     }
     
--- a/packages/external-db-config/lib/readers/aws_config_reader.js
+++ b/packages/external-db-config/lib/readers/aws_config_reader.js
@@ -1,8 +1,7 @@
 const { SecretsManagerClient, GetSecretValueCommand } = require('@aws-sdk/client-secrets-manager')
 const { checkRequiredKeys } = require('../utils/config_utils')
 
-const EmptyAWSConfig = { host: '', username: '', password: '', DB: '', SECRET_KEY: '', error: true }
-const EmptyConfig = { error: true }
+const emptyConfig = (err) => ({ secretMangerError: err.message })
 
 class AwsConfigReader {
   constructor(secretId, region) {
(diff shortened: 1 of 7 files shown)


packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -23,77 +21,85 @@ class AwsConfigReader {
     return JSON.parse(data.SecretString)
   }
 
+  async getExternalAndLocalEnvs() {
```
MXPOL, line 24: Not sure about the name, it's not external envs, it's external config variables

packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -23,77 +21,85 @@ class AwsConfigReader {
...
+
   async validate() {
-    try {
-      const cfg = await this.readExternalConfig()
-      return { missingRequiredSecretsKeys: checkRequiredKeys(cfg, ['host', 'username', 'password', 'DB', 'SECRET_KEY']) }
-    } catch (err) {
-      return { configReadError: err.message, missingRequiredSecretsKeys: [] }
-    }
+      const { config, secretMangerError } = await this.getExternalAndLocalEnvs()
+      return { missingRequiredSecretsKeys: checkRequiredKeys(config, ['host', 'username', 'password', 'DB', 'SECRET_KEY']), secretMangerError }
```
MXPOL, line 33: If we are already editing the file, maybe it will be a good idea to create a new variable for the `RequiredKeys` array.

packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -23,77 +21,85 @@ class AwsConfigReader {
     return JSON.parse(data.SecretString)
   }
 
+  async getExternalAndLocalEnvs() { 
+    const externalConfig = await this.readExternalConfig().catch(emptyConfig)
```
MXPOL, line 25: It will be much clearer, if `readExternalConfig` will **always** return an object with the next 2 properties:  `externalConfig` and `secretMangerError`

MXPOL: Therefore, it will be necessary to move the scenario in which the reading fails into the function

packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -23,77 +21,85 @@ class AwsConfigReader {
...
-        
-      } catch (err) {
-        return { configReadError: err.message, missingRequiredSecretsKeys: ['SECRET_KEY'] }
-      }
+      return { config, secretMangerError: externalConfig.secretMangerError }
     }
+
+    async readExternalConfig() {
+    const client = new SecretsManagerClient({ region: this.region })
+    const data = await client.send(new GetSecretValueCommand({ SecretId: this.secretId }))
```
MXPOL, line 61: Put it in a try & catch scope and return an empty config in case of failure with the error

packages/external-db-config/lib/readers/aws_config_reader.js
```diff
@@ -23,77 +21,85 @@ class AwsConfigReader {
...
+    return JSON.parse(data.SecretString)
```
MXPOL, line 62: If the secret reading was successful, return an object with a `externalConfig` property

PR #235, opened 2022-03-21
Skip unsupported tests

MXPOL:


packages/velo-external-db/test/e2e/app_data.e2e.spec.js
```diff
@@ -26,7 +26,7 @@ describe(`Velo External DB Data REST API: ${currentDbImplementationName()}`,  ()
         await dbTeardown()
     }, 20000)
 
-    test('find api', async() => testIfSchemaSupportsFindWithSort(env, async() => {
+    testIfSchemaSupportsFindWithSort(env)('find api', async() => {
```
Idokah, line 29: after exposing the supportedOperations const, you can pass it here instead the all env.

Idokah:
LGTM, 
I think you can (If you at all agree that it is right), change to `testIfSupportedOperationsIncludes([])` in this PR. 
and in other PR, remove the supported operations from schema providers and the tests for it

PR #242, opened 2022-03-28
Add error messages

Idokah:
Add error messages for MySQL, Postgres, MSSQL, Spanner for item already exists (trying to insert with the same _id twice).
I'll add to the rest soon.


packages/external-db-mssql/lib/sql_exception_translator.js
```diff
@@ -7,14 +7,16 @@ const notThrowingTranslateErrorCodes = err => {
...
             case 2705:
                 return new FieldAlreadyExists('Collection already has a field with the same name')
+            case 2627: 
+                return new ItemAlreadyExists(`Item already exists: ${err.message}`)
             case 4924:
                 return new FieldDoesNotExist('Collection does not contain a field with this name')
             case 2714:
                 return new CollectionAlreadyExists('Collection already exists')
             default:
                 console.log(err.number)
```
MXPOL, line 17: It will be deleted?

packages/external-db-mssql/lib/sql_exception_translator.js
```diff
@@ -7,14 +7,16 @@ const notThrowingTranslateErrorCodes = err => {
...
                 console.log(err)
```
MXPOL, line 18: In my opinion, these prints are unnecessary, because the error middleware prints these errors.

PR #246, opened 2022-03-30
Docker image for each db type

MXPOL:
This PR lets create a docker image for each db type, with the environment variable corresponding to the selected database type.

For example, if the user wants to run a cloud run instance with a Postgres database, he will use the next image link in the creation process of the cloud run instance: `gcr.io/wix-velo-api/velo-external-db:postgres`
**without the need to insert an environment variable that specifies the db type**

In addition, this pr has a special GitHub workflow that will create for each supported db an image and will push it to GCP, AWS and docker hub. 

 Interesting fact: In this way, the image size is reduced to ~200 MB compared to 570 MB in the full version

### DIFF
--- a/.github/workflows/publish-slim-version.yaml
+++ b/.github/workflows/publish-slim-version.yaml
@@ -1,0 +1,105 @@
+name: Publish Slim Images
+
+on:
+  workflow_dispatch:
+    inputs:
+      gcp:
+        type: boolean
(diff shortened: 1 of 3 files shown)


Idokah:
Looks great! 
@evgolsh

Idokah (approved):

evgolsh:
looks great indeed!

PR #276, opened 2022-05-18
Schema hooks

Idokah:


packages/velo-external-db-core/lib/router.js
```diff
@@ -26,27 +28,36 @@ const initServices = (_schemaAwareDataService, _schemaService, _operationService
     filterTransformer = _filterTransformer
     aggregationTransformer = _aggregationTransformer
     roleAuthorizationService = _roleAuthorizationService
-    hooks = _hooks || {}
+    dataHooks = _hooks?.dataHooks || {}
+    schemaHooks = _hooks?.schemaHooks || {}
```
giladwix, line 32: This exact logic is done in `app.js`, seems like an unnecessary duplication.
I would remove the change you made in `app.js` and keep this one

packages/velo-external-db-core/lib/router.js
```diff
@@ -9,10 +9,12 @@ const { secretKeyAuthMiddleware } = require('./web/auth-middleware')
...
 const path = require('path')
-const { hooksForAction, Operations, payloadFor, Actions, requestContextFor } = require('./hooks_utils')
-const { FIND, INSERT, BULK_INSERT, UPDATE, BULK_UPDATE, REMOVE, BULK_REMOVE, AGGREGATE, COUNT, GET } = Operations
+const { hooksForAction: dataHooksForAction, Operations: dataOperations, payloadFor: dataPayloadFor, Actions: DataActions, requestContextFor } = require('./data_hooks_utils')
+const { hooksForAction: schemaHooksForAction, Operations: SchemaOperations, payloadFor: schemaPayloadFor, Actions: SchemaActions } = require ('./schema_hooks_utils')
 
-let schemaService, operationService, externalDbConfigClient, schemaAwareDataService, cfg, filterTransformer, aggregationTransformer, roleAuthorizationService, hooks
+const { FIND, INSERT, BULK_INSERT, UPDATE, BULK_UPDATE, REMOVE, BULK_REMOVE, AGGREGATE, COUNT, GET } = dataOperations
+
+let schemaService, operationService, externalDbConfigClient, schemaAwareDataService, cfg, filterTransformer, aggregationTransformer, roleAuthorizationService, dataHooks, schemaHooks
```
giladwix, line 17: This line is very long and non-readable IMO.
I would do (at least) one of the following:
1. Convert `router.js` into a class, and set all of those as members
2. Unite all these variables into a class called `RouterOptions` (or any better name you can think of)

packages/velo-external-db-core/lib/router.js
```diff
@@ -9,10 +9,12 @@ const { secretKeyAuthMiddleware } = require('./web/auth-middleware')
 const { authRoleMiddleware } = require('./web/auth-role-middleware')
 const { unless, includes } = require('./web/middleware-support')
 const path = require('path')
-const { hooksForAction, Operations, payloadFor, Actions, requestContextFor } = require('./hooks_utils')
-const { FIND, INSERT, BULK_INSERT, UPDATE, BULK_UPDATE, REMOVE, BULK_REMOVE, AGGREGATE, COUNT, GET } = Operations
+const { hooksForAction: dataHooksForAction, Operations: dataOperations, payloadFor: dataPayloadFor, Actions: DataActions, requestContextFor } = require('./data_hooks_utils')
+const { hooksForAction: schemaHooksForAction, Operations: SchemaOperations, payloadFor: schemaPayloadFor, Actions: SchemaActions } = require ('./schema_hooks_utils')
```
giladwix, line 13: I would drop the multiple renames, and do the following:
Don't deconstruct the imported members from `data_hooks_utils` and `schema_hooks_utils`, and just do something like `const dataHooksUtils = require("./data-hooks-utils")`, and then for example you can do something like `dataHooksUtils.Operations`/`schemaHooksUtils.Operations`.

packages/velo-external-db-core/lib/router.js
```diff
@@ -26,27 +28,36 @@ const initServices = (_schemaAwareDataService, _schemaService, _operationService
     filterTransformer = _filterTransformer
     aggregationTransformer = _aggregationTransformer
     roleAuthorizationService = _roleAuthorizationService
-    hooks = _hooks || {}
+    dataHooks = _hooks?.dataHooks || {}
+    schemaHooks = _hooks?.schemaHooks || {}
 }
 
-const serviceContext = {
+const serviceContext = () => ({
```
giladwix, line 35: I think since it's a function you better rename this to something with a verb, like `getServiceContext`

packages/velo-external-db-core/lib/router.js
```diff
@@ -26,27 +28,36 @@ const initServices = (_schemaAwareDataService, _schemaService, _operationService
...
     schemaAwareDataService,
     schemaService
-}
+})
 
 
-const executeHooksFor = async(action, payload, requestContext) => {
+const executeDataHooksFor = async(action, payload, requestContext) => {
+    let result = payload
```
giladwix, line 42: IIUC, I would be more descriptive and call it something like `lastHookResult`

giladwix, line 42: 1. This line is bug prone: since `payload` is being passed by ref, if you change `result`(which you do) - `payload` will change as well.

2. In general, IMO `let` is a bad practice, because it makes it harder to understand what's in this variable in any point in time. a better way of doing this is to use `reduce` and setting the variable to a `const`



3. I would be more descriptive and call it something like `lastHookResult`

packages/velo-external-db-core/lib/router.js
```diff
@@ -26,27 +28,36 @@ const initServices = (_schemaAwareDataService, _schemaService, _operationService
...
+
+const executeSchemaHooksFor = async(action, payload, requestContext) => {
     let result = payload
-    for (const hook of hooksForAction(action)) {
-        result = await executeHook(hook, result, requestContext)
+    for (const hook of schemaHooksForAction(action)) {
+        result = await executeHook(schemaHooks, hook, result, requestContext)
     }
     return result
 }
```
giladwix, line 56: This is an old code, but my comments on `executeDataHooksFor` applies here

packages/velo-external-db-core/lib/router.js
```diff
@@ -243,60 +254,83 @@ const createRouter = () => {
...
     router.post('/schemas/list', async(req, res, next) => {
         try {
+            await executeSchemaHooksFor(SchemaActions.BeforeList, schemaPayloadFor(SchemaOperations.LIST, req.body), requestContextFor(SchemaOperations.LIST, req.body))
             const data = await schemaService.list()
-            res.json(data)
+            
+            const dataAfterAction = await executeSchemaHooksFor(SchemaActions.AfterList, data, requestContextFor(SchemaOperations.LIST, req.body))
+            res.json(dataAfterAction)
         } catch (e) {
+            console.log(e)
```
giladwix, line 263: Is that intentional?

packages/velo-external-db-core/lib/router.js
```diff
@@ -243,60 +254,83 @@ const createRouter = () => {
...
-            const { collectionName } = req.body
+            const collectionName  = await executeSchemaHooksFor(SchemaActions.BeforeCreate, schemaPayloadFor(SchemaOperations.CREATE, req.body), requestContextFor(SchemaOperations.CREATE, req.body))
             const data = await schemaService.create(collectionName)
-            res.json(data)
+            
+            const dataAfterAction = await executeSchemaHooksFor(SchemaActions.AfterCreate, data, requestContextFor(SchemaOperations.CREATE, req.body))
+
+            res.json(dataAfterAction)
         } catch (e) {
+            console.log(e)
```
giladwix, line 304: Also here

packages/velo-external-db-core/lib/router.js
```diff
@@ -243,60 +254,83 @@ const createRouter = () => {
...
+            const column =  await executeSchemaHooksFor(SchemaActions.BeforeColumnAdd, schemaPayloadFor(SchemaOperations.COLUMN_ADD, req.body), requestContextFor(SchemaOperations.COLUMN_ADD, req.body))
+            
             const data = await schemaService.addColumn(collectionName, column)
-            res.json(data)
+
+            const dataAfterAction = await executeSchemaHooksFor(SchemaActions.AfterColumnAdd, data, requestContextFor(SchemaOperations.COLUMN_ADD, req.body))
+
+            res.json(dataAfterAction)
         } catch (e) {
+            console.log(e)
```
giladwix, line 320: and here

packages/velo-external-db-core/lib/router.js
```diff
@@ -243,60 +254,83 @@ const createRouter = () => {
...
+            const columnName =  await executeSchemaHooksFor(SchemaActions.BeforeColumnRemove, schemaPayloadFor(SchemaOperations.COLUMN_REMOVE, req.body), requestContextFor(SchemaOperations.COLUMN_REMOVE, req.body))
+            
             const data = await schemaService.removeColumn(collectionName, columnName)
-            res.json(data)
+
+            const dataAfterAction = await executeSchemaHooksFor(SchemaActions.AfterColumnRemove, data, requestContextFor(SchemaOperations.COLUMN_REMOVE, req.body))
+            res.json(dataAfterAction)
         } catch (e) {
             next(e)
         }
```
giladwix, line 336: In general, you have a lot of unnecessary code duplication in this file. 
I understand there should be a route per operation, but the handlers can be united. I think you should think about a `routerHandler` function(name TBD of course), which gets the operation handler, before hook and after hook, and will perform the shared logic.
Alternatively you can create a class as well if you think it's more fitting to your solution

packages/velo-external-db-core/lib/schema_hooks_utils.js
```diff
@@ -0,0 +1,68 @@
+
+const hooksForAction = (action) => {
+    const hooksForAction = {
```
giladwix, line 3: This shouldn't be created each time `hooksForAction` is called, this should be a `const HOOKS_FOR_ACTION` in the top of the file.

packages/velo-external-db-core/lib/schema_hooks_utils.js
```diff
@@ -0,0 +1,68 @@
+
+const hooksForAction = (action) => {
```
giladwix, line 2: I would rename to `getHooksForAction`.
Alternatively, you can drop this function altogether, export the `HOOKS_FOR_ACTION` const I mentioned in my next comment, and just do `HOOKS_FOR_ACTION[action]`

packages/velo-external-db-core/lib/schema_hooks_utils.js
```diff
@@ -0,0 +1,68 @@
...
+        afterCreate: ['afterAll', 'afterWrite', 'afterCreate'],
+        beforeColumnAdd: ['beforeAll', 'beforeWrite', 'beforeColumnAdd'],
+        afterColumnAdd: ['afterAll', 'afterWrite', 'afterColumnAdd'],
+        beforeColumnRemove: ['beforeAll', 'beforeWrite', 'beforeColumnRemove'],
+        afterColumnRemove: ['afterAll', 'afterWrite', 'afterColumnRemove']
+    }
+    return hooksForAction[action]
+}
+
+const Operations = {
```
giladwix, line 20: wdyt about creating a `consts.js` file with all the consts, and importing them when needed?

packages/velo-external-db-core/lib/schema_hooks_utils.js
```diff
@@ -0,0 +1,68 @@
...
+    LIST: 'list',
+    LIST_HEADERS: 'listHeaders',
+    FIND: 'find',
+    CREATE: 'create',
+    COLUMN_ADD: 'columnAdd',
+    COLUMN_REMOVE: 'columnRemove',
+}
+
+const Actions = {
+    BeforeList: 'beforeList',
```
giladwix, line 30: I think you should think of a convention. Here the keys are in UpperCamelCase and in `Operations` they are in CONST_CASE. I would align

packages/velo-external-db-core/lib/router.js
Idokah: I agree, but this is the structure of router for a long time and I think this change deserves separate PR.

packages/velo-external-db-core/lib/router.js
Idokah: 2. I agree, but the reduce function does not work well with sequential asynchronous actions. If you have a more elegant solution, I would be glad to hear it.

giladwix:
@chendelbari @ZaxYonatan fyi

packages/velo-external-db-core/lib/router.js
Idokah: 1,3 fixed

packages/velo-external-db-core/lib/router.js
Idokah: I'm not sure that I understand what your alternative is.
Could you describe how the `routerHandler` would look and how that would prevent the duplication?

packages/velo-external-db-core/lib/data_hooks_utils.spec.js
```diff
@@ -3,71 +3,71 @@ const Chance = require('chance')
 const chance = Chance()
 const { Uninitialized } = require('test-commons')
 const { givenBodyWith } = require('../../velo-external-db/test/drivers/hooks_test_support')
```
MXPOL, line 5: @Idokah  We decided that we would not import things from other packages like that.......

PR #280, opened 2022-05-30
enable app info without require ejs or perform enableAppInfo

Idokah:
Until now, if a user requires the two packages (core, some-enginge-connector), he must also install `ejs` and run the command `externalDbRouter.enableAppInfo(app)` in order to render the app-info page.

Using this PR, we render the `ejs` file in the router and send the page as HTML, so the user would not be required to install `ejs` or do anything else.

Please ignore the `index.ejs` changes for now.
Let me know if you have more elegant way to read the file once ([getAppInfoTemplate](https://github.com/wix/velo-external-db/blob/3344df4f7c363d26180fb93df3637b3cbc76d0f1/packages/velo-external-db-core/lib/router.js#L22-L29))

### DIFF
--- a/packages/velo-external-db-core/lib/router.js
+++ b/packages/velo-external-db-core/lib/router.js
@@ -1,32 +1,22 @@
+const path = require('path')
+const Promise = require('bluebird')
 const express = require('express')
+const compression = require('compression')
 const { errorMiddleware } = require('./web/error-middleware')
 const { appInfoFor } = require('./health/app_info')
 const { InvalidRequest, ItemNotFound } = require('@wix-velo/velo-external-db-commons').errors
 const { extractRole } = require('./web/auth-role-middleware')
 const { config } = require('./roles-config.json')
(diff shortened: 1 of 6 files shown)


packages/velo-external-db-core/lib/router.js
```diff
@@ -17,8 +19,14 @@ const { Find: FIND, Insert: INSERT, BulkInsert: BULK_INSERT, Update: UPDATE, Bul
 
 let schemaService, operationService, externalDbConfigClient, schemaAwareDataService, cfg, filterTransformer, aggregationTransformer, roleAuthorizationService, dataHooks, schemaHooks
 
-let appInfoEnabled = false
-const enableAppInfo = () => appInfoEnabled = true
+let appInfoTemplate
+
+const getAppInfoTemplate = async() => {
```
MXPOL, line 24: I dont think that this is the right place for the function, lets move it to a util file

PR #283, opened 2022-06-09
Commons to ts

Idokah:
convert test-commons, external-db-commons to TS.


libs/test-commons/.eslintrc.json
```diff
@@ -3,15 +3,15 @@
...
+      "files": ["*.ts", "*.tsx", "*.ts", "*.tsx"],
       "rules": {}
     },
     {
       "files": ["*.ts", "*.tsx"],
       "rules": {}
     },
     {
-      "files": ["*.js", "*.jsx"],
+      "files": ["*.ts", "*.tsx"],
```
MXPOL, line 14: There are already rules for TS files above, and you set `lintFilePatterns` to be only TS files (look in the packe.json file), so rules for JS are unnecessary

libs/test-commons/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './libs/test-commons';
+import * as gen from './libs/gen';
+
+export { gen }
```
MXPOL, line 4: I think that `export * as gen from './libs/gen'`  will be nicer

libs/test-commons/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './libs/test-commons';
```
MXPOL, line 1: Semicolon??

PR #290, opened 2022-06-16
Mysql to ts

Idokah:


libs/external-db-mysql/src/index.ts
```diff
@@ -0,0 +1,29 @@
+export {default as SchemaProvider} from './mysql_schema_provider'
```
MXPOL, line 1: can you add space between the brackets? (one after and one before) like you did in line 2 for example

libs/external-db-mysql/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+export { default as DataProvider } from './mysql_data_provider'
+export { default as FilterParser } from './sql_filter_transformer'
+export {default as SchemaColumnTranslator} from './sql_schema_translator'
```
MXPOL, line 4: same here

libs/external-db-mysql/project.json
```diff
@@ -32,14 +32,14 @@
...
-        "lintFilePatterns": [ "libs/external-db-mysql/**/*.js" ]
+        "lintFilePatterns": [ "libs/external-db-mysql/**/*.ts" ]
       }
     },
     "test": {
       "executor": "@nrwl/jest:jest",
       "outputs": [ "coverage/libs/external-db-mysql" ],
       "options": {
-        "jestConfig": "libs/external-db-mysql/jest.config.js",
+        "jestConfig": "libs/external-db-mysql/jest.config.ts",
```
MXPOL, line 42: Not sure if it makes sense to rename the file to ts, if the content is written there in js (the config in the file is exported via module.exports)

libs/velo-external-db-types/src/index.ts
```diff
@@ -91,7 +91,7 @@ export type AdapterAggregation = {
 export interface IDataProvider {
```
MXPOL, line 91: IMHO, a genaric pool type should be here, so the `IDataProvider` should look like this: `IDataProvider<T>`
and then add to the this iterface the must have variables for example pool - each DataProvider must have a pool variable and the type of this variable will be T (the pool type)

libs/external-db-mysql/src/mysql_operations.ts
```diff
@@ -0,0 +1,17 @@
+import { IDatabaseOperations } from '@wix-velo/velo-external-db-types'
+import { promisify } from 'util'
+import { notThrowingTranslateErrorCodes } from './sql_exception_translator'
+
+export default class DatabaseOperations implements IDatabaseOperations {
+    pool: any
```
MXPOL, line 6: Why not specify the type of pool?

libs/external-db-mysql/src/mysql_operations.ts
```diff
@@ -0,0 +1,17 @@
...
+    query: any
```
MXPOL, line 7: maybe `MySqlQuery` type rather than `any`?

libs/external-db-mysql/src/mysql_utils.ts
```diff
@@ -0,0 +1,29 @@
+import { escapeId } from 'mysql'
+import { errors, patchDateTime } from '@wix-velo/velo-external-db-commons'
+import { Item } from '@wix-velo/velo-external-db-types'
+
+export const wildCardWith = (n: number, char: any) => Array(n).fill(char, 0, n).join(', ')
```
MXPOL, line 5: `char` variable is always a string, isn't?

libs/external-db-mysql/src/sql_filter_transformer.spec.ts
```diff
@@ -407,18 +408,36 @@ describe('Sql Parser', () => {
...
         anotherFieldName: Uninitialized,
         moreFieldName: Uninitialized,
         filter: Uninitialized,
         anotherFilter: Uninitialized,
+        anotherValue: Uninitialized,
+        moreValue: Uninitialized
+    }
+
+    interface Enviorment {
+        filterParser: any
```
MXPOL, line 437: why is this an `any` type?

libs/external-db-mysql/src/sql_filter_transformer.ts
```diff
@@ -190,20 +204,20 @@ class FilterParser {
...
 
-    extractFilterExprAndParams(havingFilter) {
+    extractFilterExprAndParams(havingFilter: { filterExpr: any; parameters: any }[]) {
         return havingFilter.map(({ filterExpr, parameters }) => ({ filterExpr: filterExpr !== '' ? `HAVING ${filterExpr}` : '',
                                                                      parameters: parameters }))
                            .concat(EmptyFilter)[0]
     }
 
-    adapterFunction2Sql(f) {
+    adapterFunction2Sql(f: any) {
```
MXPOL, line 220: isn't a `string` type?

libs/external-db-mysql/src/sql_filter_transformer.ts
```diff
@@ -220,9 +234,7 @@ class FilterParser {
         }
     }
 
-    selectFieldsFor(projection) {
+    selectFieldsFor(projection: any[]) {
```
MXPOL, line 237: why `any`?

libs/external-db-mysql/src/sql_filter_transformer.spec.ts
Idokah: Because I made interface for filter parser only for the functions that used in data provider and not to all the class.

libs/external-db-mysql/src/connection_provider.ts
```diff
@@ -1,11 +1,14 @@
...
+import SchemaProvider from './mysql_schema_provider'
+import DataProvider from './mysql_data_provider'
+import FilterParser from './sql_filter_transformer'
+import DatabaseOperations from './mysql_operations'
+import { DbProviders } from '@wix-velo/velo-external-db-types'
+import { MySqlConfig } from './types'
+
+
+export default (cfg: MySqlConfig, _poolOptions: {}): DbProviders<mysql.Pool>  => {
+    const config: mysql.PoolConfig = {
```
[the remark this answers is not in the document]
Idokah: Unfortunately, in this case it doesn't and there's an error if line 26 if I don't specify the config type.

libs/external-db-mysql/src/mysql_operations.ts
Idokah: Because of the promisify that's not the type

PR #296, opened 2022-06-21
Postgres ts migration

MXPOL:
This PR migrates the `postgres` package to TS


libs/external-db-postgres/src/connection_provider.ts
```diff
@@ -32,7 +32,12 @@ const init = (cfg, _poolOptions) => {
     const dataProvider = new DataProvider(pool, filterParser)
     const schemaProvider = new SchemaProvider(pool)
 
-    return { dataProvider: dataProvider, schemaProvider: schemaProvider, databaseOperations, connection: pool, cleanup: async() => await pool.end(() => { }) }
+    return { 
+        dataProvider: dataProvider,
+        schemaProvider: schemaProvider,
+        databaseOperations, 
+        connection: pool, 
+        cleanup: async() => pool.end(() => {})
```
Idokah, line 40: why you removed the await?

libs/external-db-postgres/src/connection_provider.ts
```diff
@@ -1,28 +1,28 @@
...
+import SchemaProvider from './postgres_schema_provider'
+import DataProvider  from './postgres_data_provider'
+import FilterParser from './sql_filter_transformer'
+import DatabaseOperations from './postgres_operations'
+import { postgresConfig } from './types'
 
 types.setTypeParser(builtins.NUMERIC, val => parseFloat(val))
 
-const init = (cfg, _poolOptions) => {
+export default (cfg: postgresConfig, _poolOptions: any) => {
```
Idokah, line 11: maybe specify that poolOptions should be some object?
{[x:string]: any}

libs/external-db-postgres/src/postgres_data_provider.ts
```diff
@@ -1,33 +1,39 @@
...
-        const resultset = await this.pool.query(`SELECT COUNT(*) AS num FROM ${escapeIdentifier(collectionName)} ${filterExpr}`, parameters)
+        const resultSet = await this.pool.query(`SELECT COUNT(*) AS num FROM ${escapeIdentifier(collectionName)} ${filterExpr}`, parameters)
                                          .catch( translateErrorCodes )
-        return parseInt(resultset.rows[0]['num'], 10)
+        return parseInt(resultSet.rows[0]['num'], 10)
     }
 
-    async insert(collectionName, items, fields) {
-        const escapedFieldsNames = fields.map( f => escapeIdentifier(f.field)).join(', ')
+    async insert(collectionName: string, items: any[], fields: any[]) {
```
Idokah, line 33: items: Item[]

libs/external-db-postgres/src/postgres_data_provider.ts
```diff
@@ -1,33 +1,39 @@
...
         const projectionExpr = this.filterParser.selectFieldsFor(projection)
-        const resultset = await this.pool.query(`SELECT ${projectionExpr} FROM ${escapeIdentifier(collectionName)} ${filterExpr} ${sortExpr} OFFSET $${offset} LIMIT $${offset + 1}`, [...parameters, skip, limit])
+        const resultSet = await this.pool.query(`SELECT ${projectionExpr} FROM ${escapeIdentifier(collectionName)} ${filterExpr} ${sortExpr} OFFSET $${offset} LIMIT $${offset + 1}`, [...parameters, skip, limit])
                                     .catch( translateErrorCodes )
-        return resultset.rows
+        return resultSet.rows
     }
 
-    async count(collectionName, filter) {
+    async count(collectionName: string, filter: AdapterFilter) {
```
Idokah, line 26: I think its better to import AdapterFilter as Filter to its a little bit confusing

libs/external-db-postgres/src/postgres_data_provider.ts
Idokah: and fields are from type ResponseField[]

libs/external-db-postgres/src/postgres_data_provider.ts
```diff
@@ -36,31 +42,32 @@ class DataProvider {
...
     }
 
-    async truncate(collectionName) {
+    async truncate(collectionName: string) {
         await this.pool.query(`TRUNCATE ${escapeIdentifier(collectionName)}`).catch( translateErrorCodes )
     }
 
-    async aggregate(collectionName, filter, aggregation) {
+    // TODO: change filter's type to Filter type
+    async aggregate(collectionName: string, filter: AnyFixMe, aggregation: Aggregation): Promise<Item[]> {
```
Idokah, line 70: why filter from Filter type isn't working here?

libs/external-db-postgres/src/sql_filter_transformer.ts
```diff
@@ -218,9 +225,11 @@ class FilterParser {
...
         }
+
+        return 
     }
 
-    orderBy(sort) {
+    orderBy(sort: Sort[]) {
         if (!Array.isArray(sort) || !sort.every(isObject)) {
             return EmptySort
         }
```
Idokah, line 235: We talked about it, but I'll write here too,
`orderBy ` purpose is to gen either Sort[] or anything else
in case of Sort[] to something and in case of anything else do something else.
Due to that, the input type is not Sort[].

libs/external-db-postgres/src/connection_provider.ts
MXPOL: image

PR #316, opened 2022-07-07
resolving get by id bug

Idokah:
As you can see [here](https://github.com/wix/velo-external-db/blob/master/libs/velo-external-db-core/src/service/data.js#L20) -  `return { item: asWixData(item[0], projection) }` - The function `packDates` treats the item as an object and collapses it if there are no items found.

I fixed it and added 2 more tests - e2e, data service test.


MXPOL (approved):
Shame on us that we didn't notice this snerio

PR #318, opened 2022-07-10
Security to ts

Idokah:


libs/external-db-security/src/services/role_authorization.ts
```diff
@@ -1,41 +1,44 @@
...
         return this.policiesFor(collectionName)?.read || DefaultPolicies
     }
 
-    writePoliciesFor(collectionName) {
+    writePoliciesFor(collectionName: string) {
         return this.policiesFor(collectionName)?.write || DefaultPolicies
     }
 
-    dataToVeloRole(role) { 
+    dataToVeloRole(role: WixDataRole): VeloRole {
```
MXPOL, line 41: Why do you need to specify exactly what result is returned from the function? TS can not deduce this alone?

MXPOL (approved):

libs/external-db-security/src/services/role_authorization.ts
```diff
@@ -1,41 +1,44 @@
+import { CollectionPermissions, VeloRole, WixDataRole } from "@wix-velo/velo-external-db-types"
```
MXPOL, line 1: can you change it to single quotes? :innocent:

Idokah: of course, btw, don't we have a lint rule for it?

PR #333, opened 2022-07-18
Skip unsupported jest each

MXPOL:
This PR adds the option to skip tests that are not supported by the database according to the supported operations

### DIFF
--- a/libs/test-commons/src/libs/test-commons.ts
+++ b/libs/test-commons/src/libs/test-commons.ts
@@ -1,4 +1,5 @@
 import { isObject } from '@wix-velo/velo-external-db-commons'
+import { SchemaOperations } from '@wix-velo/velo-external-db-types'
 
 export const Uninitialized: any = null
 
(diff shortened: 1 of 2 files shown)


libs/test-commons/src/libs/test-commons.ts
```diff
@@ -9,3 +11,12 @@ export const shouldRunOnlyOn = (impl: string[], current: string) => impl.include
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 //@ts-ignore
 export const testIfSupportedOperationsIncludes = (supportedOperations: string[], operation: string[]): any => operation.every((o: any) => supportedOperations.includes(o)) ? test : test.skip 
+
+export const testSupportedOperations = (supportedOperations: string[], arrayTable: any[][]): string[][] => {
```
Idokah, line 15: Maybe instead of string[] 
replace it to: `supportedOperations: SchemaOperations[]`

libs/test-commons/src/libs/test-commons.ts
```diff
@@ -9,3 +12,10 @@ export const shouldRunOnlyOn = (impl: string[], current: string) => impl.include
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 //@ts-ignore
 export const testIfSupportedOperationsIncludes = (supportedOperations: string[], operation: string[]): any => operation.every((o: any) => supportedOperations.includes(o)) ? test : test.skip
```
Idokah, line 14: I know tgat you didn't change this code, but maybe on the way change the supportedOperations, operation to SchemaOperations type

Idokah: And if you have a way to remove the ts-ignore from there it would be great as well  😅

Idokah (approved):

PR #352, opened 2022-10-25
Managed adapter

michaelir:


libs/velo-external-db-core/src/router.ts
```diff
@@ -94,173 +110,156 @@ export const createRouter = () => {
...
+        res.json(capabilitiesResponse)
+    })
+
     router.post('/provision', async(req, res) => {
         const { type, vendor } = cfg
-        res.json({ type, vendor, protocolVersion: 2 })
+        res.json({ type, vendor, protocolVersion: 3 })
+    })
+
+    router.get('/id', async(req, res) => {
```
MXPOL, line 128: This endpoint shouldn't be `/info` isn't it?
as written in the [profo file](https://github.com/wix-private/cloud/blob/cd837848a02f1faaa9931e06996e2b4ffe3ada93/data/cloud-data-spi/src/main/proto/com/wixpress/cloud/datasource/api/provisioning.proto#L17)

libs/external-db-spanner/src/spanner_config_validator.ts
```diff
@@ -5,7 +5,7 @@ export class ConfigValidator implements IConfigValidator {
     requiredKeys: string[]
     config: any
     constructor(config: any) {
-        this.requiredKeys = ['projectId', 'instanceId', 'databaseId', 'secretKey']
+        this.requiredKeys = ['projectId', 'instanceId', 'databaseId', 'externalDatabaseId', 'allowedMetasites']
```
Idokah, line 8: @rosenvered, I noticed it just now, 
There's a reason that only in spanner config validator specifically there is `externalDatabaseId, allowedMetasites` ? 
there was `secretKey`, but it was by mistake and I fixed it on master.

libs/external-db-mysql/src/mysql_schema_provider.ts
```diff
@@ -47,35 +50,59 @@ export default class SchemaProvider implements ISchemaProvider {
...
+        const fields = res.map(r => ({ field: r.Field, type: r.Type })).map(this.appendAdditionalRowDetails.bind(this))
+        return {
+            id: collectionName,
+            fields: fields as ResponseField[],
+            capabilities: this.collectionCapabilities(res.map(f => f.Field))
+        }
     }
 
-    translateDbTypes(row: ResponseField): ResponseField {
+    private appendAdditionalRowDetails(row: ResponseField) {
```
Idokah, line 95: This function doesn't get as a parameter `RespondField`, it returns it. @MXPOL

Idokah:
Just noticed that there's no e2e test for `upsert`. @michaelir

libs/external-db-mysql/src/mysql_schema_provider.ts
MXPOL: We be fixed in #396

PR #353, opened 2022-11-02
implementing the new data spi

michaelir:


Idokah:
IMO, the `data_source.ts` file should be under `velo-external-db-types`, so that these types can be used in other packages that are not dependent on `velo-external-db-core` (such as types itself.)

libs/velo-external-db-core/src/service/data.ts
```diff
@@ -9,13 +9,14 @@ export default class DataService {
...
+        const items = await this.storage.find(collectionName, _filter, sort, skip, limit, projection)
+        const totalCount = omitTotalCount? undefined : await this.storage.count(collectionName, _filter)
+
         return {
-            items: (await items).map(item => asWixData(item, projection)),
-            totalCount: await totalCount
-        }
+            items: items.map(item => asWixData(item, projection)),
+            totalCount: totalCount
+        }
```
Idokah, line 19: will be better to put the await in the return statement, to allow the two calls run simultaneously

libs/velo-external-db-core/src/service/data.ts
```diff
@@ -34,6 +35,11 @@ export default class DataService {
         return { item: asWixData(resp.items[0]) }
     }
 
+    async bulkUpsert(collectionName: string, items: Item[], fields?: ResponseField[]) {
+        const resp = await this.storage.insert(collectionName, items, fields, true)
```
Idokah, line 39: resp is not used

Idokah:
need to add upsert e2e test.

Idokah:
need to add aggregation with sort e2e test

libs/velo-external-db-core/src/router.ts
```diff
@@ -225,12 +193,15 @@ export const createRouter = () => {
...
             const removeRequest: dataSource.RemoveRequest = req.body;
             const collectionName = removeRequest.collectionId
-            
-            const objectsBeforeRemove = await removeRequest.itemIds.map(id => schemaAwareDataService.getById(collectionName, id))
+            const idEqExpression = removeRequest.itemIds.map(itemId => ({_id: {$eq: itemId}}))
+            const filter = {$or: idEqExpression}
+
+            const objectsBeforeRemove = (await schemaAwareDataService.find(collectionName, filterTransformer.transform(filter), undefined, 0, removeRequest.itemIds.length)).items
+
             const data = await schemaAwareDataService.bulkDelete(collectionName, removeRequest.itemIds)
```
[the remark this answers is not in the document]
michaelir: After discussing it with Max it seems like it is best to return it from the data provider itself. Since some databases like MongoDB can return the items they delete it is most efficient to use this ability. If we implement it in the outer layers we will have to do queries in order to find the items before their removal because we can't know if the data provider supports it or not. We will implement it this way in the future.

PR #354, opened 2022-11-08
New Collection SPI implementation

MXPOL:
This PR implements the new Collection SPI


libs/velo-external-db-core/src/spi-model/collection.ts
```diff
@@ -0,0 +1,165 @@
+
```
Idokah, line 1: Not really important, but if you can delete the extra lines here, and put a line between every type/interface in this file.

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -61,12 +63,135 @@ export default class SchemaService {
...
+        const collections = (!collectionIds || collectionIds.length === 0) ? 
+            await this.storage.list() : 
+            await Promise.all(collectionIds.map(async(collectionName: string) => ({ id: collectionName, fields: await this.schemaInformation.schemaFieldsFor(collectionName) })))
+                
+        const capabilities = this.formatCollectionCapabilities(this.storage.capabilities())
+        const collectionAfterFormat: Collection[] = collections.map((collection) => ({
+            id: collection.id,
+            fields: this.formatFields(collection.fields),
+            capabilities
+        }))
```
Idokah, line 89: Maybe export all of this to another private method?
And test that method as well.
(`DomainCollectionToSpi` ?)

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -61,12 +63,135 @@ export default class SchemaService {
...
+
+        return { collection: collectionAfterFormat }
+    }
+
+    formatFields(fields: ResponseField[]): Field[] {
```
Idokah, line 94: I think that all the methods that shouldn't be used from outside should be private.
(I know we didn't use that before, but maybe it's a good time to start)

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -61,12 +63,135 @@ export default class SchemaService {
...
+            type: convertFieldTypeToEnum(field.type),
+            capabilities: getFieldCapabilities(field.type)
+        }))
+    }
+
+    formatCollectionCapabilities(capabilities: DbCapabilities): CollectionCapabilities {
+        return {
+            dataOperations: capabilities.dataOperations as unknown as DataOperation[],
+            fieldTypes: capabilities.fieldTypes as unknown as FieldType[],
+            collectionOperations: capabilities.collectionOperations as unknown as CollectionOperation[],
```
Idokah, line 120: maybe create enums to `dataOperations`, `fieldTypes`, `collectionOperations` for the domain (even if it would be the same as the SPI), it would be more accurate than string and in these lines it would be more readable.

libs/velo-external-db-core/src/utils/schema_utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+}
+
+export const convertWixFormatFieldToInputFields = (field: Field): InputField => ({
+    name: field.key,
+    type: convertEnumToFieldType(field.type),
+    subtype: subtypeToFieldType(field.type)
+})
+
+export const convertWixFormatFieldsToInputFields = (fields: Field[]): InputField[] => fields.map(convertWixFormatFieldToInputFields)
```
Idokah, line 123: I think it would be more understandable - `spiFieldsToDomainFields`, or if there's a difference between input field and output Field from some reason so  - `spiFieldsToDomainInputFields`

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -61,12 +63,135 @@ export default class SchemaService {
...
+
+        const {
+            columnsToAdd,
+            columnsToRemove,
+            columnsToChangeType
+        } = this.compareColumnsInDbAndRequest(collectionColumnsInDb, collectionColumnsInRequest)
+
+        // Adding columns
+        await Promise.all(columnsToAdd.map(async(field) => await this.storage.addColumn(collection.id, field)))
+        // Removing columns
```
Idokah, line 189: new line before this

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -71,20 +72,235 @@ describe('Schema Service', () => {
...
+                type: convertFieldTypeToEnum(ctx.column.type),
+            }]
+
+            driver.givenAllSchemaOperations()
+            schema.expectSchemaRefresh()            
+            driver.givenFindResults([ { id: ctx.collectionName, fields: [] } ])
+
+            await env.schemaService.updateCollection({ id: ctx.collectionName, fields: newFields })
+
+            expect(driver.schemaProvider.addColumn).toBeCalledTimes(1)
```
Idokah, line 193: not enough to check that have been called, need to check that added the right column.
`toBeCalledWith(collectionName, SpiFieldToDomainField(ctx.field)`

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -71,20 +72,235 @@ describe('Schema Service', () => {
...
+
+            driver.givenAllSchemaOperations()
+            schema.expectSchemaRefresh()            
+            driver.givenFindResults([ { id: ctx.collectionName, fields: [] } ])
+
+            await env.schemaService.updateCollection({ id: ctx.collectionName, fields: newFields })
+
+            expect(driver.schemaProvider.addColumn).toBeCalledTimes(1)    
+            expect(driver.schemaProvider.removeColumn).toBeCalledTimes(0)    
+            expect(driver.schemaProvider.changeColumnType).toBeCalledTimes(0)
```
Idokah, line 195: only style, but I think it would be nicer instead of `.toBeCalledTimes(0)` - `.not.toBeCalled()`

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -71,20 +72,235 @@ describe('Schema Service', () => {
...
+                type: convertFieldTypeToEnum(ctx.anotherColumn.type)
+            }
+            expect(compareColumnsInDbAndRequest([], []).columnsToAdd).toEqual([])
+            expect(compareColumnsInDbAndRequest(columnsInDb, columnsInRequest).columnsToAdd).toEqual([])
+            expect(compareColumnsInDbAndRequest(columnsInDb, []).columnsToAdd).toEqual([])
+            expect(compareColumnsInDbAndRequest([], columnsInRequest).columnsToAdd).toEqual(convertWixFormatFieldsToInputFields(columnsInRequest))
+            expect(compareColumnsInDbAndRequest(columnsInDb, [...columnsInRequest, newColumn]).columnsToAdd).toEqual(convertWixFormatFieldsToInputFields([newColumn]))
+        })
+
+        test('compareColumnsInDbAndRequest function - remove columns', async() => {
```
Idokah, line 137: Under the `describe('Collection new SPI', () => {`
add another describe for all the private methods and converters.
It will look neater.

libs/velo-external-db-core/src/utils/schema_utils.ts
```diff
@@ -0,0 +1,123 @@
+import { AdapterOperators } from '@wix-velo/velo-external-db-commons'
+import { InputField, ResponseField } from '@wix-velo/velo-external-db-types'
+import { Field, FieldType, QueryOperator } from '../spi-model/collection'
+const { eq, ne, string_contains, string_begins, string_ends, gt, gte, lt, lte, include } = AdapterOperators
+
```
Idokah, line 5: I'm not sure that file is needed, maybe all of them should be private methods under the schema service.
but either way, I think you should create unit tests for all of these functions.

libs/external-db-mysql/src/mysql_schema_provider.ts
```diff
@@ -78,4 +85,51 @@ export default class SchemaProvider implements ISchemaProvider {
...
+                    columnQueryOperators: [eq, ne]
+                }
+            case 'datetime':
+                return {
+                    sortable: true,
+                    columnQueryOperators: [eq, ne, gt, gte, lt, lte]
+                }
+
+            default:
+                throw new Error(`${columnType} - Unsupported field type`)
```
Idokah, line 124: I think all that switch-case should be in another file somewhere (even utils, but I'm not sure) and should be tested.

Idokah (commented):
If you've decided to keep all the v2 logic (`schema`) 
I think maybe you should split the collection to other files (collection.ts for service, collection.e2e) because it's very confusing right now.

libs/external-db-mysql/src/mysql_schema_provider.ts
MXPOL: maybe it should be in the mysql_capabilities file - a file that will be found in every db implementation that lists the capabilities of the db in one file.

libs/velo-external-db-core/src/service/schema.ts
MXPOL: This is what prepares to be in the next PR ...

libs/velo-external-db-core/src/service/schema.ts
MXPOL: Currently can't do it, because `formatFields` function uses the storage variable, and I'm not sure if it's "right" to use the storage outside a class.

Idokah: I didn't tell to move it outside the class, only to another private method in the class

MXPOL: oh, so its done

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -11,62 +32,122 @@ export default class SchemaService {
...
+        if (columnsToRemove.length > 0) {
+            await this.validateOperation(RemoveColumn)
+        }
+        await Promise.all(columnsToRemove.map(async(fieldName) => await this.storage.removeColumn(collection.id, fieldName)))
+
+        // Changing columns type
+        if (columnsToChangeType.length > 0) {
+            await this.validateOperation(ChangeColumnType)
+        }
+        await Promise.all(columnsToChangeType.map(async(field) => await this.storage.changeColumnType!(collection.id, field)))
```
Idokah, line 78: it will be better to change it to :
`this.storage.changeColumnType?.(collection.id, field)))`
until all the implementations would support this method.

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -1,90 +1,285 @@
...
+    
+            await expect(env.schemaService.create({ id: ctx.collectionName, fields })).resolves.toEqual({
+                collection: { id: ctx.collectionName, fields }
+            })
+        })
 
-        await expect( env.schemaService.find(ctx.dbsWithIdColumn.map((db: { id: any }) => db.id)) ).resolves.toEqual( schemasListFor(ctx.dbsWithIdColumn, AllSchemaOperations) )
-    })
+        // TODO: move it to schema utils tets
+        test('compareColumnsInDbAndRequest function - add columns', async() => {
```
Idokah, line 67: I think you should choose -  move it to `schema.utils.spec` or move this function to be private method in schema service.

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -1,90 +1,285 @@
...
+            driver.givenAllSchemaOperations()
+            schema.expectSchemaRefresh()            
+            driver.givenFindResults([ { id: ctx.collectionName, fields: [] } ])
 
-        await expect( env.schemaService.list() ).resolves.toEqual( schemasWithReadOnlyCapabilitiesFor(ctx.dbsWithoutIdColumn) )
-    })
+            await env.schemaService.update({ id: ctx.collectionName, fields: newFields })
+
+
+            expect(driver.schemaProvider.addColumn).toBeCalledTimes(1)
```
Idokah, line 143: I don't know if it's necessary, but if you think you do, you might test it in the other tests as well.

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -1,90 +1,285 @@
 import * as Chance from 'chance'
 import SchemaService from './schema'
-import { AllSchemaOperations, errors } from '@wix-velo/velo-external-db-commons'
+import { errors } from '@wix-velo/velo-external-db-commons'
```
Idokah, line 3: please move up the imports from other packages

libs/velo-external-db-types/src/index.ts
```diff
@@ -140,16 +141,59 @@ export type InputField = FieldAttributes & { name: string }
 
 export type ResponseField = FieldAttributes & { field: string }
 
+export type ColumnCapabilities = {
```
Idokah, line 144: You asked me to move all the index types to another file, maybe it would be good idea here as well?

libs/external-db-mysql/src/mysql_schema_provider.ts
```diff
@@ -78,4 +85,16 @@ export default class SchemaProvider implements ISchemaProvider {
...
+        return columnCapabilitiesFor(columnType)
+    }
+
+    capabilities(): DbCapabilities {
+        return {
+            dataOperations: [],
+            fieldTypes: [],
+            collectionOperations: [],
+        }
+    }
```
Idokah, line 99: I assume it will change in the next PR?

apps/velo-external-db/test/e2e/app_schema.e2e.spec.ts
```diff
@@ -23,43 +25,61 @@ describe(`Velo External DB Schema REST API: ${currentDbImplementationName()}`,
     afterAll(async() => {
         await dbTeardown()
     }, 20000)
+    
+    describe('Velo External DB Collections REST API',  () => {
```
Idokah, line 29: Another e2e test for change type needed

libs/external-db-mysql/src/mysql_schema_provider.ts
MXPOL: Indeed

libs/velo-external-db-core/src/service/schema.spec.ts
MXPOL: I did this to make sure that addColumn is called once and not more.

Idokah (approved):

Idokah (approved):

PR #357, opened 2022-11-13
Handling bit type and default type for unkown datatype

### LINKED ISSUE #350: Unsupported column type handling
Add support for bit type in mssql.
Throw appropriate error on unknown type (or use other behaviour)

Idokah:
Fixes #350 
support bit column type.
If there's an unknown column type, return it as text.

### DIFF
--- a/libs/external-db-mysql/src/sql_schema_translator.spec.ts
+++ b/libs/external-db-mysql/src/sql_schema_translator.spec.ts
@@ -124,9 +124,13 @@ describe('Sql Schema Column Translator', () => {
 
         describe('other fields', () => {
             test('boolean', () => {
-                ['tinyint'].forEach(t => {
+                ['tinyint', 'bit'].forEach(t => {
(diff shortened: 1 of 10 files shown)


libs/external-db-mysql/src/sql_schema_translator.spec.ts
```diff
@@ -124,10 +124,14 @@ describe('Sql Schema Column Translator', () => {
 
         describe('other fields', () => {
             test('boolean', () => {
-                ['tinyint'].forEach(t => {
+                ['tinyint', 'bit'].forEach(t => {
```
MXPOL, line 127: `tinyint` to boolean - why?

Idokah: Not part of this PR but you're right, I'll fix it

MXPOL (approved):

PR #360, opened 2022-11-21
Managed adapter auth scheme

rosenvered:


Idokah:
Please change the base to managed-adapter instead of master

libs/velo-external-db-core/src/web/jwt-auth-middleware.ts
```diff
@@ -0,0 +1,71 @@
...
+        } catch (err) {
+            this.publicKey = await this.wixDataFacade.getPublicKey(this.externalDatabaseId)
+            return verify(token, this.publicKey as Secret)
+        }
+    }
+
+    async verify(token: string) {
+        const { iss, metasite } = await this.verifyWithRetry(token) as { iss: string, metasite: string }
+
+        if (iss === undefined || iss !== TOKEN_ISSUER) {
```
Idokah, line 51: I think it should be `as JwtPayload `. 
I don't think it's right to declare it as a string and then check if it's undefined. JwtPayload mark these as optional, and then it makes more sense.

Idokah: if TOKEN_ISSUER is a constant so checking if `iss` is undefined is redundant

libs/external-db-testkit/src/lib/auth_test_support.ts
```diff
@@ -1,39 +1,58 @@
...
+}
+
+const TOKEN_ISSUER = 'wix-data.wix.com'
+
+function createJwtHeader(): string {
+    const token = jwt.sign({ iss: TOKEN_ISSUER, metasite: allowedMetasite }, decodeBase64(authPrivateKey), { algorithm: 'RS256' })
+    return `Bearer ${token}`
+}
+
+function decodeBase64(data: string): string {
```
Idokah, line 38: this function exists somewhere else as well, can it be imported from there?

libs/velo-external-db-core/src/web/jwt-auth-middleware.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    test('should throw when JWT issuer is not Wix-Data', async() => {
+        const token = jwt.sign({ iss: chance.word(), metasite: ctx.metasite }, decodeBase64(ctx.authPrivateKey), { algorithm: 'RS256' })
+        await env.auth(driver.requestBodyWith(ctx.ownerRole, ctx.dataPath, `Bearer ${token}`), Uninitialized, ctx.next)
+
+        expect(ctx.next).toHaveBeenCalledWith(new UnauthorizedError('You are not authorized'))
+    })
+
+    test('should throw when JWT has no issuer', async() => {
+        const token = jwt.sign({ metasite: ctx.metasite }, decodeBase64(ctx.authPrivateKey), { algorithm: 'RS256' })
+        await env.auth(driver.requestBodyWith(ctx.ownerRole, ctx.dataPath, `Bearer ${token}`), Uninitialized, ctx.next)
```
Idokah, line 77: I think it would be clearer to send here undefined/null instead of Uninitialized

apps/velo-external-db/src/app.ts
```diff
@@ -26,22 +27,24 @@ const initConnector = async(hooks?: Hooks) => {
...
+            wixDataBaseUrl: wixDataBaseUrl || 'www.wixapis.com/wix-data'
         },
         hooks
     })
 
     return { externalDbRouter, cleanup: async() => await cleanup(), schemaProvider: providers.schemaProvider }
 }
 
-export const createApp = async() => {
+export const createApp = (wixDataBaseUrl?: string) => async() => {
```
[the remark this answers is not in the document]
MXPOL: Yes, The `E2EResources` class in the tests expects an init function, before this change the variables in the `createApp` were constants, `createApp` was the init function, but now the `wixDataBaseUrl` can be changed, so we need to create an unique init function.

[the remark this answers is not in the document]
Idokah: `const createAppWithWixDataBaseUrl = createApp.bind(null, wixDataBaseUrl())`
and send `createAppWithWixDataBaseUrl` seems to me like better solution

apps/velo-external-db/test/drivers/wix_data_testkit.ts
```diff
@@ -0,0 +1,31 @@
+import { authConfig } from '@wix-velo/test-commons'
+const express = require('express')
```
MXPOL, line 2: It is recommended to change the import to the following: 
`import express = require('express')`
That way, we won't have to provide types for the functions arguments in `app.use` 
TS will infer them by itself because now it knows what is the express variable

Idokah: `import * as express from 'express'
`
That's the way we importing it in the router, I believe it's the same.

libs/velo-external-db-core/src/web/wix_data_facade.ts
```diff
@@ -0,0 +1,33 @@
+import { errors } from '@wix-velo/velo-external-db-commons'
+const { UnauthorizedError } = errors
+import axios from 'axios'
+import { decodeBase64 } from '../utils/base64_utils'
+
+type PublicKeyResponse = {
+    publicKey: string;
+};
+
+export interface WixDataFacade {
```
MXPOL, line 10: Our convention is that Interfaces to classes should have the same name as the class only with i at the beginning
So the interface for the `WixDataFacade` class should be named `IWixDataFacade`

libs/velo-external-db-core/test/drivers/auth_middleware_test_support.ts
```diff
@@ -1,9 +1,36 @@
...
         requestContext: {
             role: role || 'OWNER',
             settings: {
-                secretKey: secretKey
-            } } } } )
+            } } },
+    header(_name: string) { return authHeader }
+} )
+
+export const signedToken = (payload: Object, expiration: string = '10000ms') =>
```
MXPOL, line 17: it's not critical, but once you gave a default value to a variable, there's no need to add its type

Idokah (approved):

MXPOL (approved):

PR #364, opened 2022-12-05
adapter-auth: change according to Wix-Data returned response

rosenvered:


libs/velo-external-db-core/src/web/jwt-auth-middleware.ts
```diff
@@ -1,14 +1,14 @@
...
+import { JwtHeader, JwtPayload, SigningKeyCallback, verify } from 'jsonwebtoken'
 import * as express from 'express'
 import { IWixDataFacade } from './wix_data_facade'
 
 
 export const TOKEN_ISSUER = 'wix-data.wix.com'
 
 export class JwtAuthenticator {
-    publicKey: string | undefined
+    publicKeys: { [key: string]: string } | undefined
```
Idokah, line 11: Can you handle undefined `allowedMetasites` (maybe with default value)

libs/velo-external-db-core/src/web/jwt-auth-middleware.ts
```diff
@@ -33,23 +33,40 @@ export class JwtAuthenticator {
         }
     }
 
+    getKey = (header: JwtHeader, callback: SigningKeyCallback) => {
```
Idokah, line 36: It would be better to use the same syntax for all the methods.

libs/velo-external-db-core/src/web/wix_data_facade.ts
```diff
@@ -4,11 +4,14 @@ import axios from 'axios'
 import { decodeBase64 } from '../utils/base64_utils'
 
 type PublicKeyResponse = {
-    publicKey: string;
+    publicKeys: {
```
Idokah, line 7: Maybe create a new type -`PublicKey` and use it instead of `{ [key: string]: string }`.
It would be clearer and more accurate to know what the functions parameters and responses are.

Idokah (approved):

PR #369, opened 2022-12-13
Fix filter by date

Idokah:


libs/test-commons/src/libs/gen.ts
```diff
@@ -22,9 +22,9 @@ export const randomEntities = (columns?: any) => {
     return arr
 }
 
-const newDate = () => {
-    const d = new Date()
-    d.setMilliseconds(0)
+const newDate = (year?: number) => {
```
MXPOL, line 25: plz add a type to `newDate`

libs/velo-external-db-core/src/converters/filter_transformer.spec.ts
```diff
@@ -104,6 +104,18 @@ describe('Filter Transformer', () => {
                 })
             })
         })
+
+    })
+
+    describe('handle filter by date', () => {
+        test('transform velo date to date object', () => {
+            const filter = {
+                [ctx.fieldName]: { $gt: genCommon.veloDate() }
```
MXPOL, line 113: can you create new 'data' property in CTX and use it here insted of generating new dates inside a test?

PR #377, opened 2022-12-19
fixes for provision flow

justinass-wix:
* renamed `/id` endpoint to `/info` and updated returned property name
* jwts signed in tests with ES256 keys as in wix-data
* keys are expected in PEM format


libs/velo-external-db-core/src/router.ts
```diff
@@ -122,12 +122,12 @@ export const createRouter = () => {
 
     router.post('/provision', async(req, res) => {
         const { type, vendor } = cfg
-        res.json({ type, vendor, protocolVersion: 3 })
+        res.json({ type, vendor, protocolVersion: 3, adapterVersion: 'v1' })
```
Idokah, line 125: What use do you have with `adapterVersion`?
Should we pass hard coded 'v1' or it should change if we're updating the adapter?

justinass-wix: It's only used in BI events. It should indicate the version of adapter code - so it could be a docker image version, or npm version.

This property has been around in v1 and v2 connectors, but I don't see it being filled... Now in v3 we marked it as mandatory.

Here's the BI event:
https://bo.wix.com/bi-catalog-webapp/#/sources/83/events/262/data-samples?artifactId=com.wixpress.cloud.cloud-data

michaelir (approved):

PR #378, opened 2022-12-19
Number type implementation using float

MXPOL:


Idokah:
can you try to disable the patch in the prepare item for insert? 
I think if you would create float column with big enough percision - (15,2) for example, it would be enough.

Idokah:
> can you try to disable the patch in the prepare item for insert? I think if you would create float column with big enough percision - (15,2) for example, it would be enough.

I tried it, it seems like that's enough, let's try to avoid for now this patch if we can, 
look at #380

MXPOL:
ל

PR #383, opened 2023-01-05
set default value for numbers to be undefined

michaelir:


libs/velo-external-db-core/src/converters/item_transformer.js
```diff
@@ -80,11 +80,11 @@ class ItemTransformer {
...
                     case 'bigint':
-                        return 0
+                        return undefined
                     case 'float':
                     case 'double':
                     case 'decimal':
-                        return 0.0
+                        return undefined
                 }
                 return 0  // default for number is 0
```
Idokah, line 89: there's another one here

PR #384, opened 2023-01-08
disable default for number insert/update

Idokah:
Change the defaults to be null for all the types but system Fields,
add two tests for insert, update numeric fields.
Fix mysql_utils bug on `patchObjectField`


.github/workflows/main.yml
```diff
@@ -44,7 +44,6 @@ jobs:
                    "mssql", "mssql17",
                    "mongo", "mongo4",
                    "firestore",
-                   "airtable",
```
MXPOL, line 47: Why did you delete this?

MXPOL (approved):

.github/workflows/main.yml
Idokah: We never really supported this, all the tests are running on some mock implementation that we maintain, so in every new test I need to add more complexity to the mock or not support this test.
Anyway, I talked with @michaelir, and we decided to let it go for now.

PR #387, opened 2023-01-11
update - from patch to fully replace

Idokah:
Delete `prepareItemForUpdate` and use `prepareItemForInsert` instead.

For collection with this schema: `_id, fieldA, fieldB`, update request with this item: 
`{_id: '..', fieldA: 'whatEver'}`, will become in `prepareItemsForInsert` to : 
`{_id: '.., fieldA: 'whatEver', fieldB: null}`

Until now, this kind of request would only update the fields that the object contains and not override the others.


michaelir (approved):

Idokah:
Will merge after merging #393 to avoid conflicts.

PR #403, opened 2023-01-31
Hide App Info option with environment variable

### LINKED ISSUE #310: Expose less information on app-info
App-info now exposes: user, host, database name.
It also exposes the permissions configuration - each table in the config and its permissions.

Idokah:
When `HIDE_APP_INFO` env not equal to 'true' the app-info will look as it looked before:
image

Otherwise, `HIDE_APP_INFO = 'true'` the config won't be sent to the HTML page, and the two sections won't be clickable.

fix #310

### DIFF
--- a/apps/velo-external-db/src/app.ts
+++ b/apps/velo-external-db/src/app.ts
@@ -21,10 +21,10 @@ const initConnector = async(hooks?: Hooks) => {
             secretKey,
             vendor,
             adapterType,
-            commonExtended: true
+            commonExtended: true,
+            hideAppInfo
         },
         hooks,
-        hideAppInfo
     })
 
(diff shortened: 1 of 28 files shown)


michaelir (approved):

apps/velo-external-db/src/app.ts
```diff
@@ -22,7 +23,8 @@ const initConnector = async(hooks?: Hooks) => {
             adapterType,
             commonExtended: true
         },
-        hooks
+        hooks,
+        hideAppInfo
```
MXPOL, line 27: I think that the right place for this variable is in the config object above

MXPOL (approved):

PR #409, opened 2023-02-06
fix generate id function for readonly collections

Idokah:
Remove the part of the code that generate _id only if it's not in the projection - because it's never in the projection if it doesn't exist.

Before:
image

After:
image


michaelir (approved):

libs/velo-external-db-core/src/service/schema_aware_data.ts
```diff
@@ -112,4 +112,10 @@ export default class SchemaAwareDataService {
         const fields = _fields ?? await this.schemaInformation.schemaFieldsFor(collectionName)
         return fields.map((f: { field: any }) => f.field)
     }
+
+    private async projectionFor(collectionName: string, _projection?: string[]) {
+        const schemaContainsId = (await this.schemaInformation.schemaFieldsFor(collectionName)).some(f => f.field === '_id')
+        const projection = _projection ?? (await this.schemaInformation.schemaFieldsFor(collectionName)).map(f => f.field)
```
MXPOL, line 118: I would create a constant to the `schemaFieldsFor` result, and then I would use it where needed insted of calling the same function twice (I know that it uses cache and not really sending queries to the DB )

libs/velo-external-db-core/src/service/schema_aware_data.spec.ts
```diff
@@ -99,7 +99,46 @@ describe ('Schema Aware Data Service', () => {
...
+        queryValidator.givenValidFilterForDefaultFieldsOf(ctx.filter) 
+        queryValidator.givenValidProjectionForDefaultFieldsOf([ctx.column.field, '_id'])
+        
+        data.givenListResult(ctx.entities, ctx.totalCount, ctx.collectionName, ctx.filter, ctx.sort, ctx.skip, ctx.limit, ['_id', ctx.column.field])
+        patcher.givenPatchedBooleanFieldsWith(ctx.patchedEntities, ctx.entities)
+
+
+        return expect(env.schemaAwareDataService.find(ctx.collectionName, ctx.filter, ctx.sort, ctx.skip, ctx.limit, [ctx.column.field])).resolves.toEqual({ 
+                                                                                                                        items: ctx.patchedEntities,
+                                                                                                              totalCount: ctx.totalCount
```
MXPOL, line 114: indent it please :)

MXPOL (approved):

PR #420, opened 2023-03-02
get port as param or env

Idokah:
in managed-adapter PR this line:
`const port = process.env.PORT || options?.port || 8080`
will become 
`const port = options?.port || 8080`

the process.env.PORT will be passed from `main.ts` and the relevant tests will pass the options object with the port,
didn't want to implement it in this PR to avoid too many conflicts, will implement it in managed-adapter


reyou:
Can we merge this?

PR #429, opened 2023-03-14
steam collection list correctly

MXPOL:


libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -28,14 +28,12 @@ export default class SchemaService {
...
-    async list(collectionIds: string[]): Promise<collectionSpi.ListCollectionsResponsePart> {        
+    async list(collectionIds: string[]): Promise<collectionSpi.Collection[]> {        
         const collections = (!collectionIds || collectionIds.length === 0) ? 
             await this.storage.list() : 
             await Promise.all(collectionIds.map((collectionName: string) => this.schemaInformation.schemaFor(collectionName)))
                 
-            return { 
-                collection: collections.map(this.formatCollection.bind(this))
-            }
+            return collections.map(this.formatCollection.bind(this))
```
Idokah, line 36:
```suggestion
            return {collections: collections.map(this.formatCollection.bind(this))}
```

I think it would be better to do this (similar to v2)

libs/velo-external-db-core/src/router.ts
```diff
@@ -276,8 +276,8 @@ export const createRouter = () => {
 
         const { collectionIds } = req.body
         try {
-            const data = await schemaService.list(collectionIds)
-            streamCollection(data.collection, res)
+            const collectionsList = await schemaService.list(collectionIds)
+            streamCollection(collectionsList, res)
```
Idokah, line 280: should stream array of 
`[{collection: Collection}, {collection: Collection}, ....]` look on query for example

Idokah (approved):

PR #453, opened 2023-09-13
Adapter with omitTotalCount flag

MXPOL:
The PR adds the option to use the `omitTotalCount` flag when making a query to the adapter. 
Please note that in order for all the tests to pass, I canceled in this PR the tests of DynamoDB, DynamoDB's SDK changed, and we will turn on again DynamoDB test in another PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,8 @@ jobs:
                    "mssql", "mssql17",
                    "mongo", "mongo4",
                    "firestore",
-                   "dynamodb",
+                  #  TODO: fix dynamodbs tests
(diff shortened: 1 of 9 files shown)


Idokah:
1. I think it's fine to build the image even though DynamoDB is broken from some reason.
2. I don't think it will improve the performance, because the `totalCount` request is in parallel with the query request. And I believe the query request is takes longer (because it's the same query that just returns more data).

Anyway, as we discussed, let's keep it as a PR, with other image URL and don't change it for everybody (you can add don't merge label to the PR)

Idokah (approved):

PR #461, opened 2023-10-04
jwt-verifier-decoder

Idokah:
Implementing the standard JWT verify & decode of SPI implementer ([see here](https://dev.wix.com/docs/rest/articles/getting-started/service-provider-interface#validating-request-signatures)).

Implemented as middleware that overrides the request body.

Still for now not using the middleware, because it would require changes in the all the e2e tests (need to encode every request), but it's ready and tested.
I think it would be better to use the middleware on another PR after merging the data SPI implementation in order to avoid too many conflicts.

what we need to do next?
1. add to config reader / config validator the publicKey, appDefId.
2. initialize the jwtVerifier with these params.
3. enable the middleware on every request + change e2es (encode every request)

### DIFF
--- a/libs/velo-external-db-core/src/web/jwt-verifier-decoder-middleware.ts
+++ b/libs/velo-external-db-core/src/web/jwt-verifier-decoder-middleware.ts
@@ -1,0 +1,22 @@
+import { Request, Response, NextFunction } from 'express'
+import { JWTVerifier } from './jwt-verifier'
+
+export class JWTVerifierDecoderMiddleware {
+    private jwtVerifier: JWTVerifier
+
+    constructor(jwtVerifier: any) {
+        this.jwtVerifier = jwtVerifier
+    }
+
+    verifyAndDecodeMiddleware() {
+        return (req: Request, res: Response, next: NextFunction) => {
+            try {
+                req.body = this.jwtVerifier.verifyAndDecode(req.body)
+                next()
+            }
+            catch (error) {
+                next(error)
+            }
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


libs/velo-external-db-core/src/web/jwt-verifier-decoder-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+            try {
+                req.body = this.jwtVerifier.verifyAndDecode(req.body)
+                next()
+            }
+            catch (error) {
+                next(error)
+            }
+        }
+    }
+}
```
MXPOL, line 22: I don't think you need a function that returns a function here, if we will use this function it as middleware we can just write `<instanceNameOfJWTVerifierDecoderMiddleware>.verifyAndDecodeMiddleware`

libs/velo-external-db-core/src/web/jwt-verifier.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    })
+})
+
+const signedToken = (payload: Record<string, unknown>, expiration = '10000ms') => signTokenWith(payload, authConfig.authPrivateKey, expiration)
+
+
+const signTokenWith = (payload: Record<string, unknown>, privateKey: string, expiration: string) => {
+    const options: jwt.SignOptions = { algorithm: 'RS256', expiresIn: expiration }
+    return jwt.sign(payload, privateKey, options)
+}
```
MXPOL, line 80: Maybe move it to test-utils file?

MXPOL (approved):
In order to continue working on the `manged-adapter` PR I approve this PR, the fixes to my comments can be made later.

PR #463, opened 2023-10-29
Collection v4 SPI support

MXPOL:


libs/external-db-spanner/src/sql_schema_translator.ts
```diff
@@ -91,12 +91,28 @@ export default class SchemaColumnTranslator implements ISpannerSchemaColumnTrans
                 return `STRING${this.parseLength(2 ** 16)}`
 
             case 'text_large':
+            case 'text_language':
                 return `STRING${this.parseLength(2 ** 32)}`
```
Idokah, line 95: `text_language` shouldn't be something like - 'en'?
How did we get this huge size?
it's right to all implementations - I saw you translated language to big texts everywhere, maybe we should reconsider.

libs/external-db-spanner/src/sql_schema_translator.ts
```diff
@@ -91,12 +91,28 @@ export default class SchemaColumnTranslator implements ISpannerSchemaColumnTrans
                 return `STRING${this.parseLength(2 ** 16)}`
```
Idokah, line 91: I know it's not related to this pr, and that you didn't set that, but maybe we should consider the size of the string types here, 
text_small = 2**8 (which is reasonable)
text_medium = 2**16 (65,536 chars) - which is not medium in any way - I think it should be something around 2**12 (4,096) 
text_large = 2**32 (59125081250218501850218590852109581 chars approx) - which is extremely large (specifically when we are talking about `text_language`), I think we should go with around 2**20 or something close to that.

libs/velo-external-db-core/src/utils/schema_utils.spec.ts
```diff
@@ -36,22 +36,33 @@ describe('Schema utils functions', () => {
...
+            [VeloFieldTypeEnum.language, FieldType.text],
+            [VeloFieldTypeEnum.richContent, FieldType.object],
+            [VeloFieldTypeEnum.mediaGallery, FieldType.object],
+            [VeloFieldTypeEnum.address, FieldType.object],
+            [VeloFieldTypeEnum.pageLink, FieldType.object],
+            [VeloFieldTypeEnum.reference, FieldType.object],
+            [VeloFieldTypeEnum.object, FieldType.object],
+            [VeloFieldTypeEnum.array, FieldType.object],
+          ])('%s type', (veloType, ourType) => {
+            expect(wixDataEnumToFieldType(veloType)).toBe(ourType)
```
Idokah, line 64: Maybe `domainType` instead of `ourType`? not critical

libs/velo-external-db-core/src/utils/schema_utils.spec.ts
```diff
@@ -60,23 +71,33 @@ describe('Schema utils functions', () => {
...
+            [VeloFieldTypeEnum.language, 'language'],
+            [VeloFieldTypeEnum.richContent, 'richContent'],
+            [VeloFieldTypeEnum.mediaGallery, 'mediaGallery'],
+            [VeloFieldTypeEnum.address, 'address'],
+            [VeloFieldTypeEnum.pageLink, 'pageLink'],
+            [VeloFieldTypeEnum.reference, 'reference'],
+            [VeloFieldTypeEnum.object, 'object'],
+            [VeloFieldTypeEnum.array, 'array'],
+          ])('%s type', (veloType, ourSubType) => {
+            expect(subtypeToFieldType(veloType)).toBe(ourSubType)
```
Idokah, line 99: same

libs/velo-external-db-core/src/web/error-middleware.ts
```diff
@@ -2,7 +2,7 @@ import { NextFunction, Response } from 'express'
 import { domainToSpiErrorTranslator } from './domain-to-spi-error-translator'
 
 export const errorMiddleware = (err: any, _req: any, res: Response, _next?: NextFunction) => {
-  if (process.env['NODE_ENV'] !== 'test') {
+  if (process.env['NODE_ENV'] === 'test') {
```
Idokah, line 5: ?

Idokah (changes requested):
Didn't look on the hooks yet? Have they changed?

libs/velo-external-db-core/src/web/error-middleware.ts
MXPOL: I encountered some issues with the CI tests, and I needed to examine the errors. However, I inadvertently left the error messages visible in the tests. Since errors should not be present in tests, I have reverted the tests to their original state.

libs/external-db-spanner/src/sql_schema_translator.ts
MXPOL: This change is not directly related to this PR, and the PR is already quite large. I suggest addressing this change in a separate PR to maintain a clear scope for this PR.

libs/external-db-spanner/src/sql_schema_translator.ts
MXPOL: I don't know how a language item will look like, expect the fact its a string type. 
But you are right I will chnage it to something smaller.

libs/external-db-spanner/src/sql_schema_translator.ts
Idokah: np, wrote it because I've already been there, you are welcome to do it in another PR, shouldn't cause any conflicts obviously

libs/velo-external-db-core/src/web/error-middleware.ts
Idokah: bottom line, if i understand correctly, it was always `===`, and you changed it some when and now you're changing it back?

MXPOL: Nope, the opposite it correct, it always was !==, that means the errors should not appers in tests, because there are tests that test errors and we don't want to see them in the tests results.. 
Changed it in `manged-adapter` and I forgot to fix this before rebasing this PR with `manged-apdater` branch ..

libs/velo-external-db-types/src/collection_types.ts
```diff
@@ -29,24 +29,35 @@ export enum FieldType {
     multiReference = 'multiReference',
 }
 
+export enum SubFieldType {
+    
+}
```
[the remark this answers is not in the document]
MXPOL: I wanted to at first. But that just makes everything harder.
So not right now....

libs/external-db-mssql/src/sql_schema_translator.ts
```diff
@@ -92,6 +92,7 @@ export default class SchemaColumnTranslator {
 
 
             case 'boolean_':
+            case 'boolean_boolean':
```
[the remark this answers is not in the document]
MXPOL: Done. I have updated the sql_schema_translator tests in each implementation.

libs/external-db-postgres/src/postgres_data_provider.ts
```diff
@@ -53,12 +53,16 @@ export default class DataProvider implements IDataProvider {
...
     }
 
-    async delete(collectionName: string, itemIds: string[]) {
+    async delete(collectionName: string, itemIds: string[]): Promise<number> {
         const rs = await this.pool.query(`DELETE FROM ${escapeIdentifier(collectionName)} WHERE _id IN (${prepareStatementVariables(itemIds.length)})`, itemIds)
                              .catch( translateErrorCodes )
+        
+        // @ts-ignore
         return rs.rowCount
     }
```
MXPOL, line 67: @Idokah  
I have no idea why, but the CI didn't want to compile it and claimed that update / delete can return an answer that is null ( you can see the errors near the failed comments)
No matter what I did, it wouldn't compile the Postgres code. 
That's why I added `@ts-ignore`

Idokah (approved):

PR #467, opened 2023-11-08
JWT config reader refactor

MXPOL:
This PR adds an option to config reader to read more 2 environment variables (`JWT_PUBLIC_KEY`, `APP_DEF_ID`)
So that later they will be used for JWT middleware 

In addition, I deleted the `EXTERNAL_DATABASE_ID` variable in the config reader that was in the previous JWT milestone.

### DIFF
--- a/libs/external-db-config/test/drivers/external_config_reader_e2e_test_support.ts
+++ b/libs/external-db-config/test/drivers/external_config_reader_e2e_test_support.ts
@@ -1,12 +1,12 @@
 
 import { Uninitialized } from '@wix-velo/test-commons'
 import { create } from '../../src/factory'
-import awsMySql = require('./aws_mysql_config_test_support')
-import awsMongo = require('./aws_mongo_config_test_support')
-import azureMySql = require('./azure_mysql_config_test_support')
-import gcpMySql = require('./gcp_mysql_config_test_support')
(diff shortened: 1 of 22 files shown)


libs/external-db-config/test/drivers/external_config_reader_e2e_test_support.ts
```diff
@@ -1,12 +1,12 @@
...
-import azureMySql = require('./azure_mysql_config_test_support')
-import gcpMySql = require('./gcp_mysql_config_test_support')
-import gcpSpanner = require('./gcp_spanner_config_test_support')
-import gcpFirestore = require('./gcp_firestore_config_test_support')
+import * as awsMySql from './aws_mysql_config_test_support'
+import * as awsMongo from './aws_mongo_config_test_support'
+import * as azureMySql from './azure_mysql_config_test_support'
+import * as gcpMySql from './gcp_mysql_config_test_support'
+import * as gcpSpanner from './gcp_spanner_config_test_support'
+import * as gcpFirestore from './gcp_firestore_config_test_support'
```
MXPOL, line 9: I already worked on this file, so I fixed the import of the files properly ...

guyrotem (approved):
evidently, adding an env var in the current design is a bit more labor-some than it could have been, but considering that it's quite an uncommon event, LGTM

PR #474, opened 2023-11-22
Schema create refactor

MXPOL:


libs/velo-external-db-core/src/utils/schema_utils.ts
```diff
@@ -207,10 +208,16 @@ export const responseFieldToWixFormat = (fields: ResponseField[]): collectionSpi
     })
 }
 
+export const fieldKeyToPrecision = (fieldKey: string): number | undefined => {
+    const fieldsWithPrecision = ['_id', '_owner']
```
Idokah, line 212:
```suggestion
    const FieldsWithPrecision = ['_id', '_owner']
```

Idokah (changes requested):
let's add another test in `schema.spec.ts`
It's true that you're testing `fieldKeyToPrecision` in `schema_utils.spec.ts` but nothing assures that you're really using this function.

on `schema.spec.ts` i expect to see a test that call create schemaService with `_id`, `_owner` and other columns and in the `expectCreateWithFieldsOf` expect to get the columns with the percision where needed.

Idokah:
as we discussed, let's check in schema.ts that we're getting _id

and another thing - not critical but I think it would be better - we can add to  `_id` field the attribute `isPrimary: true` (in the same way we are adding the percision) and then in each provider handle the isPrimary. 
from the same reason that you're not setting the percision of _id/_owner in each provider by the column name, isPrimary has the same logic

libs/velo-external-db-core/src/utils/schema_utils.spec.ts
```diff
@@ -96,20 +98,38 @@ describe('Schema utils functions', () => {
...
+                isPrimary: false,
+            })
+        })
+        test.each(FieldsWithPrecision)('convert %s field in velo format to our fields with precision property', (columnName) => {
+            expect(wixFormatFieldToInputFields({ key: columnName, type: fieldTypeToWixDataEnum('text') })).toEqual({
+                name: columnName,
+                type: 'text',
+                subtype: 'string',
+                precision: 255,
+                isPrimary: columnName === '_id' ? true : false,
```
Idokah, line 132: constant

libs/velo-external-db-core/src/service/schema.spec.ts
```diff
@@ -31,36 +33,39 @@ describe('Schema Service', () => {
...
-            }]
+            const fields = [
+                { key: PrimaryKeyFieldName, type: VeloFieldTypeEnum.text },
+                InputFieldToWixFormatField(ctx.column),
+            ]
+
             driver.givenAllSchemaOperations()
             schema.expectSchemaRefresh()            
-            driver.expectCreateWithFieldsOf(ctx.collectionName, fields)
+            driver.expectCreateWithFieldsOf(ctx.collectionName, WixFormatFieldsToInputFields(fields))
```
Idokah, line 60: lets add here test that we are calling the schema with the right columns.
` expect(driver.schemaProvider.create).toBeCalledWith(collectionName, columnsAfterTranslation)`

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -100,6 +101,17 @@ export default class SchemaService {
...
+        if (!fields.length) {
+            throw new errors.InvalidRequest('Fields are missing')
+        }
+
+        const fieldsName = fields.map(field => field.key)
+        if (!fieldsName.includes(PrimaryKeyFieldName)) {
+            throw new errors.InvalidRequest(`${PrimaryKeyFieldName} is field is missing`)
+        }
+    }
+
```
Idokah, line 114: 1. it's not async.
2. the first check is redundant - it would succeed only if there's _id column, so you don't need to check if it's empty before it.

libs/velo-external-db-core/test/drivers/schema_information_test_support.ts
```diff
@@ -14,16 +16,42 @@ export const givenDefaultSchemaFor = (collectionName: any) => {
 export const expectSchemaRefresh = () =>
     when(schemaInformation.refresh).mockResolvedValue(undefined)
 
-export const givenSchemaFor = (collectionName: string, fields: {field: string, type: string, subtype?: string}[]) => {
+export const givenSchemaFieldsFor = (collectionName: string, fields: {field: string, type: string, subtype?: string}[]) => {
     when(schemaInformation.schemaFieldsFor).calledWith(collectionName)
                                             .mockResolvedValue( fields.map(field => ({ field: field.field, type: 'string' })) )
 }
 
+export const givenschemaFor = (collectionName: string, fields: ResponseField[], capabilities: any) => {
```
Idokah, line 24:
```suggestion
export const givenSchemaFor = (collectionName: string, fields: ResponseField[], capabilities: any) => {
```

libs/velo-external-db-core/src/service/schema.ts
Idokah: and maybe `${PrimaryKeyFieldName} is field is missing` --> `${PrimaryKeyFieldName} field is missing`

Idokah (approved):

libs/velo-external-db-core/src/service/schema.ts
```diff
@@ -33,10 +33,11 @@ export default class SchemaService {
             await this.storage.list() : 
             await Promise.all(collectionIds.map((collectionName: string) => this.schemaInformation.schemaFor(collectionName)))
 
-        return { collections: collections.map(this.formatCollection.bind(this)) }
+                return { collections: collections.map(this.formatCollection.bind(this)) }
     }
 
-    async create(collection: collectionSpi.Collection): Promise<collectionSpi.CreateCollectionResponse> {                
+    async create(collection: collectionSpi.Collection): Promise<collectionSpi.CreateCollectionResponse> {   
+        await this.validateFields(collection.fields)
```
Idokah, line 40: if its not async you can remove await :) 
thats the last comment promise - you can ignore as well, bye

PR #487, opened 2024-01-02
Collection change not supported error

MXPOL:
I created a new error in our domain for CollectionChangeNotSupported. At the moment, the implementation is only for MySQL and Postgres

### DIFF
--- a/libs/external-db-postgres/src/postgres_operations.ts
+++ b/libs/external-db-postgres/src/postgres_operations.ts
@@ -12,6 +12,6 @@ export default class DatabaseOperations implements IDatabaseOperations {
     // @ts-ignore
     async validateConnection() {
         return await this.pool.query('SELECT 1').then(() => { return { valid: true } })
-                              .catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e) } })
+                              .catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e, '') } })
     }
 }
--- a/libs/velo-external-db-commons/src/libs/errors.ts
+++ b/libs/velo-external-db-commons/src/libs/errors.ts
@@ -1,4 +1,4 @@
-class BaseHttpError extends Error {
+export class BaseHttpError extends Error {
     constructor(message: string) {
         super(message)
     }
@@ -14,7 +14,7 @@ export class CollectionDoesNotExists extends BaseHttpError {
     collectionName: string
     constructor(message: string, collectionName?: string) {
         super(message)
-        this.collectionName = collectionName || ''
+        this.collectionName = collectionName ?? '' 
     }
 }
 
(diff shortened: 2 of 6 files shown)


libs/external-db-postgres/src/postgres_operations.ts
```diff
@@ -12,6 +12,6 @@ export default class DatabaseOperations implements IDatabaseOperations {
     // @ts-ignore
     async validateConnection() {
         return await this.pool.query('SELECT 1').then(() => { return { valid: true } })
-                              .catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e) } })
+                              .catch((e) => { return { valid: false, error: notThrowingTranslateErrorCodes(e, '') } })
```
Idokah, line 15: is it needed? can we make the parameter optional and transform it to empty string if needed in `notThrowingTranslateErrorCodes` ?

libs/velo-external-db-commons/src/libs/errors.ts
```diff
@@ -120,3 +120,14 @@ export class UnrecognizedError extends BaseHttpError {
...
 }
+
+export class CollectionChangeNotSupportedError extends BaseHttpError {
+    collectionName: string
+    fieldName: string
+
+    constructor(message: string, collectionName?: string, fieldName?: string) {
+        super(message)
+        this.collectionName = collectionName || ''
+        this.fieldName = fieldName || ''
```
Idokah, line 131: I think it would be better to use ?? instead to handle null/undefined.
because theoretically || could catch "falsy" values as well (I think there's no real impact here unless someone will call is table - 'undefined' / 'null' / 'false'), but still I think ?? is describes better what you're trying to do

Idokah (changes requested):
lets add e2e to that in `app_schema.e2e.spec.ts`  (see in app_data.e2e how we did it, add another section of error handling, and check it only on mysql/postgres.)
It would give us 
1. check that its really working
2. some "documentation" that describes that it's implemented only for mysql and postgres.

libs/external-db-postgres/src/postgres_operations.ts
MXPOL: I need the collection name for the error in changing the column type, and I'd rather it not be optional, so I must put an empty string for the collection name argument. 
We did the same thing in MySQL. I can set a default value for the collection name argument ...., is it better?

libs/velo-external-db-commons/src/libs/errors.ts
MXPOL: No problem. I will change it to all the errors that have it

libs/external-db-postgres/src/postgres_operations.ts
Idokah: If you need the collection name only for part of the errors, it sounds like an optional param, putting default value is fine as well

MXPOL:
@Idokah  Regarding the response of the e2e tests, it will be in the next PR, which translates the error from our domain to Wix's domain.

Idokah (approved):

PR #496, opened 2024-01-15
External db logger

MXPOL:


libs/velo-external-db-core/src/index.ts
```diff
@@ -65,6 +77,7 @@ export class ExternalDbRouter {
 
     isInitialized(connector: DbConnector) {
         if (!connector.initialized) {
+            this.logger.error('Connector must be initialized before being used')
             throw new Error('Connector must be initialized before being used')
```
Idokah, line 81: maybe we should catch the error in some middleware (as I think we already do) and log them there, I don't think we want to log before every throw

Idokah (commented):
Can you elaborate a little why are we using third party dep for logging? what its has that console doesn't? 
I do agree that we should add some logs in lower level in order to debug more easily when needed, I just not convinced yet that we need all of this for it. we are not sending these logs anywhere and don't have plans to do so in the near future. 

I think for this point we can only add environment variable with the debug level and add logs wherever we need with console, as for now the way for us the monitor is through Wix-data anyway.

libs/external-db-mysql/src/index.ts
```diff
@@ -9,20 +9,22 @@ export * as testResources from '../tests/e2e-testkit/mysql_resources'
...
 import init from './connection_provider'
+import { Logger } from '@wix-velo/external-db-logger'
 
 export const driver = () => require('../tests/drivers/sql_filter_transformer_test_support')
 export const opsDriver = () => require('../tests/drivers/db_operations_test_support')
 
 
 export class MySqlConnector extends DbConnector {
-    constructor() {
+    constructor(logger: Logger) {
```
guyrotem, line 19: maybe use ILogger to enable providing alternative implementations?

libs/external-db-logger/src/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+        },
+      })
+  }
+
+  debug(message: string, data?: Data): void {
+    this.logger.debug({ data }, message)
+  }
+
+  info(message: string, data?: Data): void {
+    this.logger.info({ data }, message)
```
guyrotem, line 26: any other global meta-data we want to add?
maybe the DC? some cloud-run meta-data the could be of interest?

libs/external-db-mysql/src/index.ts
```diff
@@ -9,20 +9,22 @@ export * as testResources from '../tests/e2e-testkit/mysql_resources'
...
         super(MySqlConfigValidator, init)
         this.type = 'mysql'
+        this.logger = logger
     }
 }
 
-export const mySqlFactory = async(config: any, options: any): Promise<DatabaseFactoryResponse> => {
-    const connector = new MySqlConnector()
-    const { connection, cleanup, ...providers } = await connector.initialize(config, options)
+export const mySqlFactory = async(config: any, logger: Logger, options: any): Promise<DatabaseFactoryResponse> => {
```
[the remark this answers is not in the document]
MXPOL: Using the same logger instance for all prints is a common practice. 
And it will allow us to define the settings of the prints in only one place instead of all console.log appearances.

libs/external-db-logger/src/logger.ts
MXPOL: Oops, I missed it
At the moment I can't think of such useful information

apps/velo-external-db/src/storage/factory.ts
```diff
@@ -1,11 +1,12 @@
+import { Logger } from '@wix-velo/external-db-logger'
 import { DatabaseFactoryResponse } from '@wix-velo/velo-external-db-commons'
 
-export const engineConnectorFor = async(_type: string, config: any): Promise<DatabaseFactoryResponse> => {
+export const engineConnectorFor = async(_type: string, config: any, logger?: Logger): Promise<DatabaseFactoryResponse> => {
```
Idokah, line 4: ILogger?

libs/external-db-mysql/src/index.ts
```diff
@@ -10,20 +10,23 @@ import { MySqlConfigValidator } from './mysql_config_validator'
...
         super(MySqlConfigValidator, init)
         this.type = 'mysql'
+        this.logger = logger
     }
 }
 
-export const mySqlFactory = async(config: MySqlConfig, options: any): Promise<DatabaseFactoryResponse> => {
-    const connector = new MySqlConnector()
-    const { connection, cleanup, ...providers } = await connector.initialize(config, options)
+export const mySqlFactory = async(config: MySqlConfig, logger: Logger, options: any): Promise<DatabaseFactoryResponse> => {
```
Idokah, line 28: ILogger?

libs/velo-external-db-core/src/index.ts
```diff
@@ -19,6 +20,14 @@ import { Router } from 'express'
 import { CollectionCapability } from './spi-model/capabilities'
 import { decodeBase64 } from './utils/base64_utils'
 
+interface ExternalDbRouterConstructorParams { 
+    connector: DbConnector;
+    config: ExternalDbRouterConfig;
+    logger?: Logger;
```
Idokah, line 26: ILogger?

libs/velo-external-db-core/src/index.ts
```diff
@@ -35,10 +44,13 @@ export class ExternalDbRouter {
     cleanup: ConnectionCleanUp
     router: Router
     config: ExternalDbRouterConfig
-    constructor({ connector, config, hooks }: { connector: DbConnector, config: ExternalDbRouterConfig, hooks: {schemaHooks?: SchemaHooks, dataHooks?: DataHooks}}) {
+    logger?: Logger
```
Idokah, line 47: ILogger?

PR #497, opened 2024-01-17
Error response fix

MXPOL:


Idokah (approved):
LGTM, lets just check it once on production to be sure it works

PR #509, opened 2024-02-04
create encrypted field as object

Idokah:
we are creating encrypted field as string instead as an object

### DIFF
--- a/libs/velo-external-db-core/src/utils/schema_utils.ts
+++ b/libs/velo-external-db-core/src/utils/schema_utils.ts
@@ -217,8 +217,9 @@ export const fieldKeyToPrecision = (fieldKey: string): number | undefined => {
     return FieldsWithPrecision.includes(fieldKey) ? 255 : undefined
 }
 
-export const wixFormatFieldToInputFields = (field: collectionSpi.Field): InputField => ( field.encrypted ? encryptedInputField(field.key) : 
-{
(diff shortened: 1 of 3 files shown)


libs/velo-external-db-core/src/utils/schema_utils.ts
```diff
@@ -217,14 +217,20 @@ export const fieldKeyToPrecision = (fieldKey: string): number | undefined => {
     return FieldsWithPrecision.includes(fieldKey) ? 255 : undefined
 }
 
-export const wixFormatFieldToInputFields = (field: collectionSpi.Field): InputField => ({
+export const wixFormatFieldToInputFields = (field: collectionSpi.Field): InputField => ( field.encrypted ? encryptedInputField(field.key) : 
+{
```
MXPOL, line 221: It took me a few moments to understand what's going on here, is it possible to extract the creation of a non-encrypted field into a function like you did for an encrypted column?

MXPOL (approved):

libs/velo-external-db-core/src/utils/schema_utils.ts
Idokah: [nonEncryptedInputField](https://github.com/wix/velo-external-db/pull/509/commits/a9ab1c7d3bdf2e08bf9e35021ba0025cc4becc6c)

PR #513, opened 2024-02-11
add default sort

Idokah:
in order to support paging the way wix-data expecting, we need to have default sort for each query.


MXPOL (approved):

MXPOL (approved):
Amazing work

PR #542, opened 2025-04-15
Req payload limit increase adapter v2

MXPOL:
[`libs/velo-external-db-core/src/router.ts`](diffhunk://#diff-f3856fdca70ec7dc347e5de8f7ea04d9c43984b6089f6d917ab8baf0103a78a7L82-R83): Updated the `express.json()` middleware to set a limit of '4mb' for incoming JSON payloads and added `express.urlencoded()` middleware with a '4mb' limit for URL-encoded payloads.


Idokah (approved):
🚀

